PR #112, opened 2014-11-14
Implemented a global mode.

pauxus:
Using the configuration parameter useGlobal (${npm.useGlobal}), no node
nor npm is installed. Instead, a globally installed instance in the path
is used. This is usefully for corporate installations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 .idea
 *.iml
 target
+.classpath
+.project
+.settings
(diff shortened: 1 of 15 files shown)


brianmgray:
What other plugins run a local npm/gulp? I like the philosophy of this plugin for backend developers on my project. I work on both the backend and front end so I have all of these tools installed locally and would prefer not to download another copy of all of them. So I like the idea of a flag I can pass in / maven profile to tell this to use the npm on my path.

eirslett:
There is no other plugin (to my knowledge) that lets you choose between running project-specific tools or system-wide tools. You'd have to make it yourself. You could of course just fork this plugin and merge the pull request, and then maintain your own fork, if that's all you need.
(That being said; disk space is cheap, and node.js/npm isn't so big in terms of megabytes...)

brianmgray:
Well with the complexities of running executables across multiple environments, I couldn't get anything else working, so indeed this plugin is my best bet. I personally don't see the harm in having it use npm if it is on the path (maybe ensuring the correct version?) but indeed after the initial download there is not much downside of having two copies on my machine. Thanks!

Issue #89, opened 2014-09-26
Provide a "skip" parameter


arienkock:
I have a parent POM defining most of the frontend-plugin's executions, but they are not needed in the POM module. I'd like to be able to add skip parameter and then re-enable it in the child projects.

willtran-:
@arienkock I have the same issue. A big multi-module project where only some modules need to run certain tasks, but all modules follow the same conventions / configuration if they do need the run that task. I've addressed that in this pull request https://github.com/eirslett/frontend-maven-plugin/pull/140. Along with the previous pull request, you can do something like this, which will install node in the projects root folder (not each module) and will only install node and run npm install if there is a src/main/resources/package.json in the module:
```
<plugin>
	<groupId>com.github.eirslett</groupId>
	<artifactId>frontend-maven-plugin</artifactId>
	<version>0.0.21-SNAPSHOT</version>
	<configuration>
		<nodeInstallDirectory>${session.executionRootDirectory}</nodeInstallDirectory>
		<workingDirectory>src/main/resources</workingDirectory>
		<ifExists>src/main/resources/package.json</ifExists>
	</configuration>
	<executions>
		<execution>
			<id>install node and npm</id>
			<goals>
				<goal>install-node-and-npm</goal>
			</goals>
			<configuration>
				<nodeVersion>v0.10.30</nodeVersion>
				<npmVersion>1.4.21</npmVersion>
			</configuration>
		</execution>
		<execution>
			<id>npm install</id>
			<goals>
				<goal>npm</goal>
			</goals>
		</execution>
	</executions>
</plugin>
```

arienkock:
First questions that come to mind are: How is the <ifExists> different from
a profile activated by presence of package.json (suggested earlier)? and if
multiple projects share the same node and module dirs and they're not
homogenous then they may conflict, and if they are, then running install
more than once is overkill.

eirslett:
Ok, we have skip flags in place now, so I'll close this issue. Please reopen it if there's some case that isn't covered!

PR #119, opened 2014-11-25
Add support for skipping execution

howama:
Add options for skipping all mojos. Particularly useful when multiple configurations are being used, e.g. several grunt executions with different targets for karma tests, protractor tests, main build, etc.

Related to https://github.com/eirslett/frontend-maven-plugin/issues/89


eirslett:
Thanks! This was merged as 4170c47.

orrgal1:
how do you use this in the pom?

eirslett:
Try something like this:

```
<configuration>
  <skip>${skipTests}</skip>
</configuration>
```

(I haven't tried that snippet myself)

PR #125, opened 2014-12-02
Execute 'npm run-script' on incremental builds as well

KnisterPeter:
This enables npm executions like build-scripts which are
configured in the package.json as scripts.

### DIFF
--- a/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NodeAndNPMInstaller.java
+++ b/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NodeAndNPMInstaller.java
@@ -144,13 +144,18 @@ final class DefaultNodeAndNPMInstaller implements NodeAndNPMInstaller {
                 // see https://github.com/eirslett/frontend-maven-plugin/issues/65#issuecomment-52024254
                 File packageDirectory = new File(config.getInstallDirectory(), "./node/package");
                 if (packageDirectory.exists() && !npmDirectory.exists()) {
-                    packageDirectory.renameTo(npmDirectory);
+                    if (! packageDirectory.renameTo(npmDirectory)) {
+                        logger.warn("Cannot rename NPM directory, making a copy.");
+                        FileUtils.copyDirectory(packageDirectory, npmDirectory);
+                    }
                 }
                 logger.info("Installed NPM locally.");
             } catch (DownloadException e) {
                 throw new InstallationException("Could not download npm", e);
             } catch (ArchiveExtractionException e) {
                 throw new InstallationException("Could not extract the npm archive", e);
+            } catch (IOException e) {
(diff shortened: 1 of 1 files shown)


cvakiitho:
:+1: , This would make a lot of things easier, at least for me :)

gyandeeps:
@KnisterPeter @eirslett any update on this?
I am waiting on a functionality where i can run any `scripts` from the `package.json` file.

KnisterPeter:
If @eirslett and @karlvr agree I would write a bigger change which incorporates the build models into my patch for the npm mojo.

eirslett:
Maybe you could outline what you're planning? I'm reluctant to add complex changes to the plugin, but if it's something that greatly improves the integration for Eclipse, it could be a welcome change. I don't use Eclipse, and @karlvr is pretty much the export on that field, so we would have to get his opinion as well!

KnisterPeter:
I would like to extend the npm-mojo in a way to trigger incremental builds
and add triggerfiles like in the gulp- and grunt-mojo. Maybe includes and
excludes so npm is only triggered for related files. Thats all which is
required.

2015-04-03 13:56 GMT+02:00 Eirik Sletteberg <notifications@github.com>:

> Maybe you could outline what you're planning? I'm reluctant to add complex
> changes to the plugin, but if it's something that greatly improves the
> integration for Eclipse, it could be a welcome change. I don't use Eclipse,
> and @karlvr <https://github.com/karlvr> is pretty much the export on that
> field, so we would have to get his opinion as well!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/eirslett/frontend-maven-plugin/pull/125#issuecomment-89266937>
> .
>



-- 
____________________________________

          Markus Wolf
          Kieler Straße 21
          22769 Hamburg

     tel: (+49) 40 / 80 600 599
     mob: (+49) 177 / 288 48 67
     web: http://www.matrixweb.de
 twitter: @KnisterPeter
____________________________________

karlvr:
@KnisterPeter Could you elaborate on what you mean by "incorporates the build models"?

KnisterPeter:
I meant by that to include "full builds" and "incremental builds".
Currently the npm module is only run on full builds. This is mostly not what you want if you have a script bock in your package json.

okitu:
This issue has never been fixed, and there is no way around it, because of this:

https://github.com/eirslett/frontend-maven-plugin/blob/c7a76cc6c66e3db901e44b555fdf90830e1b3709/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/NpmMojo.java#L60

And at the same time gulp, grunt & webpack builds (I don't know about the others) do not correctly refresh the workspace in incremental  eclipse builds.

Maybe adding the `triggerFiles` and `srcDir` configuration to npm builds, or adding support for incremental builds, when the npm arguments are not "install".

Thank you for looking into this!

PR #139, opened 2015-01-21
Task only runs <ifExists>somefile</ifExists>

willtran-:
Added configuration parameter <ifExists> to all tasks, where execution
of a task only occurs if that file exists. Addresses [Issue
#89](https://github.com/eirslett/frontend-maven-plugin/issues/89)


willtran-:
Ooops, inclides a previous unrelated commit.

Issue #143, opened 2015-01-21
Possible bug


cowwoc:
Why are both `result` and `error` pointing to stdout? https://github.com/eirslett/frontend-maven-plugin/blob/master/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ProcessExecutor.java#L40

Shouldn't `error` point to stderr? Is this a bug?

eirslett:
This was fixed, thanks for the PR! I'll close this issue then.

PR #145, opened 2015-01-23
Error string was being read from stdout instead of stderr #143

cowwoc:


cowwoc:
Seems like a configuration error with appveyor, not the pull request.

eirslett:
Looks better now. Thanks :-)

Issue #148, opened 2015-02-02
nonProxyHosts ignored


andham:
It seems as the nonProxyHosts config is ignored for the npmDownloadRoot in the install-node-and-npm mojo. Possibly the same goes for nodeDownloadRoot, but I haven't tried.

andham:
Just checked the code and the nonProxyHosts config for Maven proxies is fully ignored in both these download cases.

eirslett:
Yup. It's not implemented.

bdellegrazie:
nonProxyHosts should not be ignored. The problem is that npm can only used nonProxyHosts if the proxy configuration is defined as an environment variables (http_proxy, https_proxy, no_proxy (recently))
There is no support for no_proxy in the .npmrc file
See: https://github.com/npm/npm/issues/7168

However what I'm not sure about is the behaviour of this plugin w.r.t. environment variables - is the execution environment passed on to NPM or is it 'cleaned'?

PR #149, opened 2015-02-02
Proxy config now includes (and displays) id

andham:
Signed-off-by: Anders Hammar <anders@hammar.net>


eirslett:
What is the ID used for?

andham:
The id is...well. an id. :-) It's used to distinguish one proxy def from others. I ran into issue #148 and displaying the id of the proxy used simplifies when you have more than one def.

eirslett:
Sorry for the delay! I'll merge this!

eirslett:
ouch, looks like there is a merge conflict?

andham:
The merge conflict should be fixed now.

eirslett:
Thanks! The build was failing, but I'm sure this change isn't the reason.

PR #150, opened 2015-02-03
Add npm to the path when executing processes

karlvr:
Some modules have post-install scripts that want to use the npm executable. I have simply added npm to the PATH environment variable used when running node.

See imagemin/imagemin#68 and kevva/bin-wrapper#35 for examples of this causing problems.

### DIFF
--- a/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ProcessExecutor.java
+++ b/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ProcessExecutor.java
@@ -85,12 +85,15 @@ final class ProcessExecutor {
                 }
             }
         }
-        if (pathVarValue == null) {
-            environment.put(pathVarName, workingDirectory + File.separator + "node");
+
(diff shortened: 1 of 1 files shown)


dcendents:
Any chance this can be merged?

I have a scenario where bower installs some lib but it needs to be compiled manually.
To do this the grunt-exec plugin is used to run "npm install -q && make" in the lib directory (to install the necessary npm dependencies as required by make).

Now, to make this work on the build server which uses the frontend plugin and does not have nodejs or npm installed, this needs to be changed to "../../node/node ../../node/npm/bin/npm-cli.js install -q && make"

I'd love to be able to keep a single configuration for grunt to run the same command, using system nodejs or through the frontend plugin.

thanks

karlvr:
@eirslett Have you had a chance to consider this? I believe this change won't break anything, and it does improve a few cases where tools run `npm` themselves.

eirslett:
Sure, I'll merge this in! Thanks for the PR, and sorry for being so late!

karlvr:
@eirslett Thank you, and no problem! Love your work.

PR #153, opened 2015-02-05
Adding support for the “ember” command line, from ember-cli

joachimhs:
In order to be able to support ember-cli builds with this plugin via Maven, I have added an EmberMojo class, and and "ember" build goal. 

I suppose there might also be an "InstallEmberCli" mojo that executed npm install ember-cli, however.


chrisvdp:
Can it run the ember cli tests via `ember test`?

joachimhs:
Running ember test is currently not built in, but this is definitely one area where we are also interested in extending the plugin into. Making the build fail when test fail is an important step towards getting Ember streamlined into the CI process.

eirslett:
Thanks! Looks good to me.
If "ember test" returns exit code 1 (or something other than 0) when it fails, the build should fail too.

chrisvdp:
@joachimhs how does it work?

PR #157, opened 2015-02-11
Maven proxy security

kwa:
Maven supports encrypted passwords for server and proxy configurations in the settings.xml (see http://maven.apache.org/guides/mini/guide-encryption.html).
This pull request makes it possible to have the encrypted password in the settings.xml.

Furthermore, when the plugin runs it logs the password in clear/plain text which I think could be improved upon so I am masking the password when the arguments are displayed in the log.

This feature will make it possible for me to use it at the  company I work for =)


eirslett:
Thanks for the PR! :-)

kwa:
Do you think it would be possible for release this version (latest version) of the plugin to be released to a public maven repository soon?
Best regards

PR #160, opened 2015-02-14
Improve ember-cli support

flecno:
* Move EmberMojo to EmberBuildMojo
 * Add EmberTestMojo
 * Add ember-cli example project
 * Add ember usage to README

@joachimhs can you review this?


flecno:
Ember-test fails in CI because ember-cli can't find phantomjs.

I don't want to install phantomjs globally so I prefer to install it local with `package.json`. Ember-CLI uses [testem](https://github.com/airportyh/testem) as testrunner. I have no idea how to add the correct path to phantomjs in the `testem.json` config. So we need to add `node_modules/.bin` to the path or use npm scripts. Ember-CLI adds a test scipt per default, which finds phantomjs (because `npm run` uses `node_modules/.bin` in the path?).

I think the npm scripts option is the right way. But I also prefer to add `node_modules/.bin` to the `PATH` when executing node at all. What's your opinion?

eirslett:
Try to fix testem! Karma uses a local version of phantomjs when running tests. I'm sure testem should be able to do the same.
Otherwise; It's a great example project, but I don't think it should live in this code repo; It deserves its own Git repo. As for adding a dedicated EmberTestMojo; what value does it provide, that you cannot easily achieve with the current EmberMojo?

flecno:
Oh, the current testem 0.7.0 pre-release supports local phantomjs (https://github.com/airportyh/testem/pull/486) and there is a open PR in ember-cli (https://github.com/ember-cli/ember-cli/pull/3285). I changed my example project to use this ember-cli PR. It seems to work on linux/unix. I have to check the windows failures.

@eirslett I agree to move the example project into other repos. But how we can use it here for "integration tests"?
And you are right; I'm going to merge the `Ember*Mojo`'s back into `EmberMojo`.

eirslett:
If you move it into a separate repo, you can add a Travis CI build to that repo to verify that things work.

The "integration tests" in this plugin aren't meant to be comprehensive; it's just to test the very core functionality, and verify that it can actually run a build. Maven builds are very slow integration tests, so I'd rather only have one. The risk, of course, is that a bug in the Maven plugin could break Ember support, but I think that such a bug would break Grunt support as well - so the existing test covers it. (Unfortunately, we don't know that in advance...)

flecno:
I moved back to the `EmberMojo`, which can be used also for tests and I'm going to extract the example project into it's own project.

I think the next steps are waiting to get the testem 0.7 PR merged into ember-cli. Then I'm going to squash my commits and it's ready to merge.

eirslett:
I'm not sure if we need to merge this PR at all? I mean, if the Ember example is moved to a separate repo, and testem is fixed so it uses a local phantomjs instance, isn't that all that is needed?

flecno:
Yes, you are right. I will publish the ember example and update the Readme next time.

PR #178, opened 2015-03-18
SLF4J done the right way

stephenc:
* [Maven 3.1.x uses SLF4J as built in logger](http://maven.apache.org/maven-logging.html)
* With Maven 3.3.1 the shim that has been used actually breaks the build.
* The shim is pointless anyway as the minimum version of Maven for this plugin is 3.1.0


eirslett:
Thanks!

Philipp91:
Will there be a frontend-maven-plugin release containing this fix, soon? The plugin does not work with the current Maven version (3.3.1).

PR #187, opened 2015-03-31
Add npm/node download root docs

chrisvdp:
functionality exists, but isn't documented.


eirslett:
Ah, thanks a lot! :-)

PR #191, opened 2015-04-03
[Fix #190] Bower doesn't pass down proxy settings

### LINKED ISSUE #190: Bower doesn't pass down proxy settings
Bower doesn't pass proxy settings from Maven's configuration down like NPM does.

busches:
Note our corporate proxy uses only http was not able to test https, but it's mirrored after the NPM settings

This fixes issue #190

### DIFF
--- a/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/BowerMojo.java
+++ b/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/BowerMojo.java
@@ -1,13 +1,17 @@
 package com.github.eirslett.maven.plugins.frontend.mojo;
 
 import com.github.eirslett.maven.plugins.frontend.lib.FrontendPluginFactory;
+import com.github.eirslett.maven.plugins.frontend.lib.ProxyConfig;
 import com.github.eirslett.maven.plugins.frontend.lib.TaskRunnerException;
+import org.apache.maven.execution.MavenSession;
 import org.apache.maven.plugin.AbstractMojo;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
+import org.apache.maven.plugins.annotations.Component;
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
+import org.apache.maven.settings.crypto.SettingsDecrypter;
 
 import java.io.File;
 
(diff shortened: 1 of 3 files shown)


eirslett:
I think this should instead be achieved with a `.bowerrc` file in the working directory:
http://bower.io/docs/config/

The intention of Maven's reusing proxy config is for downloading Node and npm itself, not for the build tools (grunt, gulp, karma, bower...). The reason is that such an approach would couple your backend build tighter to your frontend build than necessary. You should be able to run bower from the command line in your project and it would behave exactly as if it were run from Maven. Which is why I suggest the .bowerrc approach. The loose coupling between the builds, and keeping frontend-maven-plugin as thin as possible (well, I admit it's already growing very think) should justify the code duplication of repeating proxy configuration.

Is that a satisfactory explanation?

busches:
The use case we're going for here is running the build on our CI machines which have to use the proxy to get out to npm/bower, additionally we want them to execute all the steps, including the build steps.

eirslett:
Cannot that be solved with .bowerrc?

kwa:
So as I mentioned in #192 I do not believe using .npmrc or .bowerrc or any other .rc file is a good solution. 

I am not sure if I am too impolite to say this as it may come across that I know better what the plugin is trying to accomplish than the author... but here goes:
Background
For me the purpose of the plugin is to integrate node/bower/grunt invocations when running a maven build. To accomplish this (normally) maven uses **_generic_** POMs which can be used by anyone as-is and this plugin is intended to be compatible with maven. So to make POMs dynamic maven uses settings.xml to be able to parametrize parts of POMs which are (normally) checked in to a version control system. The parameters can be user specific and located in a place each user can modify without having to modify the source code of the current maven project. (Of course if you have a property in a POM which can only be found using a settings.xml file then your build is not portable and so on).

My views:
Maven uses the settings.xml so I as a user (of the current maven project I am trying to build) do not have to change anything inside the project, and this is especially true for for example `<server>` and `<proxy>` where a person would be hard pressed to come up with a proxy which fits everybody using the current project (think dev teams in different countries). This is in fact a decoupling of the project from user specific needs to make it more reusable.
As @busches points out it is common to build maven project on CI servers which has a "maven" way of providing build specific settings: the POMs and/or the settings.xml. If we use .npmrc and others we would have to have versions of all these on the CI server as well... but (hope not too impolite) this is the whole purpose of the plugin - integrate maven way of doing things to be able to use bower/node and so on.
Also. I am assuming that people use the plugin because they have one or many "larger" maven projects where the node/front-end  is simply another sub-part, thus files like .rc files are checked in into version control. If the plugin helps/centralizes the configuration to one place instead of many, I think that would be a good thing.

From what I have seen in the code it seems that would be possible to let all xxxRunners have access to the proxy configuration and I would be willing to help with that if there are wishes for that.

Well these are my most salient points about why settings.xml for a maven plugin is better option than .rc files are.

YHuard:
@kwa I get your point and I agree with you. IMHO the main benefit of using this maven plugin is (as written in the Readme) to let "backend developers can run a clean build without even installing Node [globally] on their computer".  Following that logic, I agree that's all proxy settings should be passed by maven when using maven, without taking care of existing .xxxrc files. But of course the frontend developper would still have to set his own settings in .xxxrc files when using Node directly.

Furthermore it's easy to pass those proxy settings to npm (`npm --proxy=...` or `npm --https-proxy=...`) and bower (`bower --config.proxy=...` or `bower --config.https-proxy=...`)

As you said, proxy settings can be specific to a team and thus are not global settings that you can share in .xxxrc files

Issue #192, opened 2015-04-07
Regex error in NodeTaskExecutor


yhuard:
Hi

I get the following error when executing goal 'npm':

```
[ERROR] Failed to execute goal com.github.eirslett:frontend-maven-plugin:0.0.23:npm (npm install) on project ****: Execution npm install of goal com.github.eirslett:frontend-maven-plugin:0.0.23:npm failed: Look-behind group does not have an obvious maximum length near index 11
[ERROR] //(?<keep>.*):.*@
```

It is related to the regex in NodeTaskExecutor.java, line 65. I'm using Java 6 and Windows 7 on this project. Could you help me?

eirslett:
Ouch... @kwa could you please have a look at this? I'm not a regex ninja, unfortunately.
@YHuard can you post a simple example file (or project) that reproduces the error?

YHuard:
I tried to run the example projet you give and got the same error. The only thing I changed is the pom.xml with the following configuration:

```xml
                <version>0.0.23</version>

                <executions>

                    <execution>
                        <id>install node and npm</id>
                        <goals>
                            <goal>install-node-and-npm</goal>
                        </goals>
                        <configuration>
                            <nodeVersion>v0.12.2</nodeVersion>
                            <npmVersion>2.7.5</npmVersion>
                        </configuration>
                    </execution>
```

To be more specific, I'm running Maven 3.2.5 (the latest 3.3.x isn't compatible with Java 6) and I use eclipse. I also set a proxy, which is probably why the regex is called

eirslett:
Yup, it's probably because of the proxy. I think you could use version 0.0.22 as well, since you use Maven 3.2. This problem was introduced in version 0.0.23. If you could provide your proxy settings (you can anonymize it of course) that would probably be very helpful!

YHuard:
My proxy is simply a raw ip and a port (`protocol='http', host='17x.xx.xxx.xx', port=8080`), there's no authentication with a password.

Version 0.0.22 works like a charm

kwa:
@YHuard so you have something like this in your settings.xml:
```
<proxies>
  <proxy>
    <id>example-proxy</id>
    <active>true</active>
    <protocol>http</protocol>
    <host>17x.xx.xxx.xx</host>
    <port>8080</port>
  </proxy>
</proxies>
```
?

It would be interesting to know if you get this problem if you run using java 7 or java 8 (I see that you are running in windows so perhaps it is not easy to change...)

[...]

PR #195, opened 2015-04-08
Fix Regex error in NodeTaskExecutor #192

kwa:
Attempt to find user name and password in proxy argument string and mask it.
Passes the following likely and some very unlikely scenarios:
```
public static class Testing {
    public static void main(String[] args) {
        assert NodeTaskExecutor.maskPassword(null) == null;
        assert "".equals(NodeTaskExecutor.maskPassword(""));
        assert " ".equals(NodeTaskExecutor.maskPassword(" "));
        assert "proxy=//abc:def@example.com:8080".equals(NodeTaskExecutor.maskPassword("proxy=//abc:def@example.com:8080"));
        assert "proxy=http://abc:***@1.2.3.4:8080".equals(NodeTaskExecutor.maskPassword("proxy=http://abc:def@1.2.3.4:8080"));
        assert "proxy=http://abc:***@example.com:8080".equals(NodeTaskExecutor.maskPassword("proxy=http://abc:de:f@example.com:8080"));
        assert "proxy=http:////:***@example.com:8080".equals(NodeTaskExecutor.maskPassword("proxy=http:////:::@example.com:8080"));
        assert "proxy=http://///@/:***@example.com:8080".equals(NodeTaskExecutor.maskPassword("proxy=http://///@/::://:@:@example.com:8080"));
        assert "proxy=http:// ///@/:***@example.com:8080".equals(NodeTaskExecutor.maskPassword("proxy=http:// ///@/::://:@ :@example.com:8080"));
        assert "proxy=https:// ///@/:***@example.com:8080".equals(NodeTaskExecutor.maskPassword("proxy=https:// ///@/::://:@ :@example.com:8080"));
    }
}
```

### DIFF
--- a/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NodeTaskExecutor.java
+++ b/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NodeTaskExecutor.java
@@ -1,16 +1,19 @@
 package com.github.eirslett.maven.plugins.frontend.lib;
 
-import static com.github.eirslett.maven.plugins.frontend.lib.Utils.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import static com.github.eirslett.maven.plugins.frontend.lib.Utils.*;
 
 abstract class NodeTaskExecutor {
+    private static final String DS = "//";
+    private static final String AT = "@";
+    
     private final Logger logger;
     private final String taskName;
     private final String taskLocation;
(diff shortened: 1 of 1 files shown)


eirslett:
I fixed the failed build now. Could you please rebase this on top of master, and then see if it builds?

kwa:
@eirslett I have updated this branch. I tried to rebase but need to merge a conflict between "Proxy config now includes (and displays) id", the getId() in ProxyConfig constructor caused the conflict. I see now that the rebase did not do what I expected. Not sure if that is why AppVeyor is having problems...

Should I redo and perform a merge instead?... It should yield a fast-forward... I will test locally.

kwa:
I did a merge instead and that worked better. Still unfortunate that the rebase did not work.. wondering if rebase and merge uses different merge strategies as I did not get any merge conflicts when using a regular merge.
Seems that the build has passed, so everthing OK or not @eirslett ?

eirslett:
Everything looks ok now! Thanks!

PR #202, opened 2015-04-20
Add Webpack support

yobert:
I've never contributed to this project before so this is more of an RFC: I'd love to add webpack support to this excellent plugin. What do you think?


philostler:
+1

**Just** found this maven plugin and was disappointed to see no webpack support... then I see this!

eirslett:
Thanks a lot, looks good to me!

Issue #204, opened 2015-04-24
Node + NPM installation directories inconsistency


NBardelot:
The project I'm working on requires Maven 2, so I'm using the 0.0.16 version of the plugin.

After Node and NPM are installed, with default properties for the plugin, I end up with directories like this:

[code]
./node_modules
./node <-- node binary is here
./node/npm
./node/npm/bin <-- npm bash file is here with npm-cli.js
[/code]

The npm bash file is:

[code]
$ cat npm
#!/bin/sh

basedir=`dirname "$0"`

case `uname` in
    *CYGWIN*) basedir=`cygpath -w "$basedir"`;;
esac

if [ -x "$basedir/node.exe" ]; then
  "$basedir/node.exe" "$basedir/node_modules/npm/bin/npm-cli.js" "$@"
else
  node "$basedir/node_modules/npm/bin/npm-cli.js" "$@"
fi
[/code]

This script is incompatible with how NodeAndNPMInstaller handles the installation.

I propose two solutions:
Solution 1 : remove this script at the end of the installation, and create an new script at the root of the project (i.e. where pom.xml stands) which will handle things properly.
Solution 2 : install NPM in a way that matches this script's behaviour.

Solution 1 has the advantage of being compatible with older versions of the plugin without having to clean up the directories, and requires less code change. But solution 2 is cleaner, and makes life easier for someone who uses other installations of node+NPM.

eirslett:
The script inside there isn't used at all from the plugin, it's calling npm via the node executable. You might have some luck using the maven clean plugin if you want to hack something together for your particular project?

NBardelot:
Thanks for this quick response Eirik. 

You plugin is very good, because it enables the team to work on a single project:
- our frontend developers is in charge of the frontend module, and they can setup a full dev environement (with node + npm) in one Maven command line,
- our backend developers can package the whole project without mastering node+npm in the frontend module.

So I understand what you say, and I have no problem with the inner behaviour of the plugin : a "mvn clean install" works fine for us. 

But What's disturbing is that after running the Maven goals the developer will have the node+npm tools installed, but will not be able to use this utility script as-is. You'd be confused if after installing Maven the mvn.bat or mvn shell scripts weren't working, wouldn't you? Well, it's the same here with the npm script. Untill it works our developers must call npm using the full node command line, which works fine but requires to setup the PATH and NODE_PATH and write the full directory structures... It would really help if the npm script worked.

eirslett:
[...]

PR #205, opened 2015-04-29
Allow automatic installation of npm helper scripts that match the directory structure

NBardelot:
See also: https://github.com/eirslett/frontend-maven-plugin/issues/204

Commit 4c0f28c makes it possible to install npm helper scripts during the installation of node+npm. I moved the script from the example project's helper_scripts directory to the core module. They're embedded as ressources and copied to the working directory if the option is set to true (it's false by default as you asked).

Commit c17e8e4 updates the project's documentation to take into account this feature.

B. R.


eirslett:
I'm sorry I haven't looked at this PR before now, but what I think is that it's better to add decent documentation on how to use the helper scripts, than to auto-generate them. That way, we don't add too much complexity to the plugin.

PR #206, opened 2015-04-29
Prepend node and npm directories to PATH

TiddoLangerak:
Appending these folders to PATH give them the lowest possible priority, thus if any other version of node/npm is found in PATH these will be used instead when called from a build script. This can lead to errors due to (e.g.) incompatible versions or permission problems.

By prepending the directories to PATH instead the bundled versions will take precedence over any other version already present on the system. This will ensure that any call to node or npm from the frontend-maven-plugin will use the bundled version, thus eliminating the problems mentioned above.

### DIFF
--- a/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ProcessExecutor.java
+++ b/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ProcessExecutor.java
@@ -87,11 +87,12 @@ final class ProcessExecutor {
         }
 
         StringBuilder pathBuilder = new StringBuilder();
-        if (pathVarValue != null) {
-            pathBuilder.append(pathVarValue).append(File.pathSeparator);
-        }
+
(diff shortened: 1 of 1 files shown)


TiddoLangerak:
On second thought, this will give problems of itself: the `npm` script in the `npm/bin` folder doesn't seem to work on it's own at the moment. Both prepending and appending will thus give problems when that script is the only version of `npm` in path. I've opened an issue for this at the npm repo, perhaps that should be solved first (see [here])(https://github.com/npm/npm/issues/8130)).

eirslett:
Have you tried commiting this file into your project root directory? (or wherever the build script is running from) https://github.com/eirslett/frontend-maven-plugin/blob/master/frontend-maven-plugin/src/it/example%20project/helper-scripts/npm

TiddoLangerak:
No, not yet. Didn't realize that there was already a helper script in the repo. But I suppose that would already fix the issues I had.

eirslett:
Ok, in that case I'll close this issue!

PR #209, opened 2015-05-05
Add plexus-utils and use CommandLineUtils to parse arguments

bkonkle:
Thanks for your continued work on this plugin!

This should solve the issue I'm having with spaces, but if you want to avoid adding a dependency you may want to look into CommandLineUtils.translateCommandline() to see how they handle it in plexus-utils.


bkonkle:
No activity on this for an extended period of time, so I'm closing.

PR #211, opened 2015-05-07
Changed testFailureIgnore property to be consistent with surefire.

pauxus:
Currently, the parameter to ignore test failures (testFailureIgnore) is mapped to the property 'testFailureIgnore'. This is inconsistent with maven surefire, which uses the property "maven.test.failure.ignore".

By using using this property as well, we will automatically get the same handling in Jenkins as with regular Java builds, i.e. the build will be marked as unstable instead of failed when test failures occur.

This PR introduces a slight usage change, because servers explicitly using "-DtestFailureIgnore" might behave differently (configuration in the POM is unaffected)

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,7 +202,7 @@ By default, no colors will be shown in the log.
 ```
 __Skipping tests:__ If you run maven with the `-DskipTests` flag, karma tests will be skipped.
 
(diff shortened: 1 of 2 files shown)


mriehema:
+1

eirslett:
Thanks!

PR #212, opened 2015-05-07
Run karma in the process-test-classes phase by default.

pauxus:
This is the last phase before test and allows karma to be run BEFORE maven-surefire-plugin in a war project.

That way, test results of karma (using a junit-reporter) can automatically be picked up by CI servers listening explicitly for the result of surefire (Jenkins and Hudson).

Otherwise, karma would run after surefire, and when Jenkins reads test result, it would not get karma result.

### DIFF
--- a/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/KarmaRunMojo.java
+++ b/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/KarmaRunMojo.java
@@ -13,7 +13,7 @@ import org.apache.maven.plugins.annotations.Parameter;
 import org.slf4j.LoggerFactory;
 
 
-@Mojo(name="karma",  defaultPhase = LifecyclePhase.TEST)
+@Mojo(name="karma",  defaultPhase = LifecyclePhase.PROCESS_TEST_CLASSES)
 public final class KarmaRunMojo extends AbstractMojo {
 
     /**


eirslett:
I think that's rather a problem with your Jenkins setup? It works fine here. Jenkins looks for files in the surefire reports folder.

pauxus:
A just double checked again (test environment, war project with frontend-plugin enabled), and no: it does not work, which matches the workings of the maven-plugin for Jenkins (SurefireArchiver.postExecute()):

- SA runs after every executed mojo
- first checks whether the current mojo is a test mojo (via the TestMojo enum)
- if not, SA simply exits, which is what happens here
  - If it IS a test mojo, SA checks if their are any new / unprocessed files in output folder and handles them

I can't imagine why it CAN work in your setup.

By applying the PR, surefire is guaranteed to run after frontend, which also SA running for surefire to just pick up the result on the way.

An alternative would be to include frontend:karma in maven-plugin's TestMojo class, which would be the more complex approach, because frontend-plugin would have to propose a reportsDirectory (which does not make sense when not using the junit-reporter).

eirslett:
I suggest you either

1) Register frontend-maven-plugin as a TestMojo here: https://github.com/jenkinsci/maven-plugin/blob/d3eceb5d62e799dc7cfdba10d92d133069d389d7/src/main/java/hudson/maven/reporters/TestMojo.java (they should accept pull requests)
or
2) Override the lifecycle phase manually in your pom.xml

I'd want tests to be run in the test phase, and rather fix problems at the root cause.

PR #219, opened 2015-05-19
npmRegistryURL

paulgallagher75:
Extra property setting 'npmRegistryURL' added so that no changes are required to the pom.xml to specify a local/internal registry.

Required because internal build policy at Red Hat has to allow repeatable builds, which means we need to mirror the NPM registry,  and as we do not want to change code (pom.xml) and add a static internal redhat NPM registry, which will not be public, this way we can specify the registry during our build process and still leave the source to use the default when external.

### DIFF
--- a/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/InstallNodeAndNpmMojo.java
+++ b/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/InstallNodeAndNpmMojo.java
@@ -27,6 +27,12 @@ public final class InstallNodeAndNpmMojo extends AbstractFrontendMojo {
     private String npmDownloadRoot;
 
     /**
+     * Registry override, passed as the registry option during npm install if set.
+     */
+    @Parameter(property = "npmRegistryURL", required = false, defaultValue = "")
+    private String npmRegistryURL;
+
+    /**
      * Where to download Node.js and NPM binaries from.
      *
      * @deprecated use {@link #nodeDownloadRoot} and {@link #npmDownloadRoot} instead, this configuration will be used only when no {@link #nodeDownloadRoot} or {@link #npmDownloadRoot} is specified.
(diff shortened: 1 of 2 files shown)


paulgallagher75:
Hi, 

I rebased my commits so its a bit more tidy so it is just the changes to add the property

eirslett:
I believe a better solution to this is to add a `~/.npmrc` file to your build server (and workstation). Then, it is also independent of Maven. And you can build third-party projects as well (that don't have a plugin version with support for this feature). Would that suffice?

paulgallagher75:
Hi,

The problem we have with the ~/.npmrc is that we use continues build servers and user and the home directory would have to be unique (jenkins would be a problem), we also use a locked down build env for the final productised releases and that would be very difficult for us to get the file in there and alter it.

We can however control the properties very easily in metadata about each build so this is a much simpler way for us to build products that use the plugin

eirslett:
The configuration could also be put in `/etc/npmrc` to get around the home directory problem. But you'd have to convince the ops team to tweak the build server (add that file), which is maybe not an option?

Isn't also `NPM_CONFIG_REGISTRY` an environment variable you can set to change the registry? From the build configuration, it should be possible to override that variable and it's passed down all the way to npm.

paulgallagher75:
Hi,

The /etc/npmrc would not work as we have an individual internal npm registry setup for every build we do (this is to cache internally so we can do repeatable builds).

The NPM_CONFIG_REGISTRY would have been a good option if it wasn't that it is very difficult to inject env settings into our product build system, and controlling the properties passed on the mvn command line is very easy for us in our systems.

eirslett:
So it's difficult to do `NPM_CONFIG_REGISTRY=foo mvn clean install`, but easy to do `mvn clean install -DnpmConfigRegistry=foo`?
Anyways, I'll merge it, it's an edge case but I guess Red Hat isn't the only restrictive organisation out there. Thanks for the PR!

paulgallagher75:
Yes thats basically it in a nutshell, internal build systems.

Thanks

PR #236, opened 2015-06-18
Clarify README to state download roots must point to the same repo

geneticgrabbag:
Regarding download roots, the install instructions state that node and npm may have different roots "as they are now stored in separate repos".  However, the example implies the repo must be the same for both roots.
This change rewords that sentence to better convey that, while the download roots may be different, the downloads must come from the same repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Node/npm will only be "installed" locally to your project. It will not be instal
 </execution>
 ```
 
-You can also specify separate download roots for npm and node as they are now stored in separate repos.
+You can also specify separate download roots for npm and node as long as they are stored in the same repo.
 ```xml
 <execution>
     ...


geneticgrabbag:
Hopefully, I inferred the intent of the README sentence based on the example.

eirslett:
The functionality is set up so that you can download node from nodejs.org and npm from npmjs.org. And I guess you would think of them as separate repositories?

geneticgrabbag:
Ah, I reread the original text and I now understand your meaning.
Maybe, the example could be change from:
```
  <nodeDownloadRoot>http://artifactory/nodejs/</nodeDownloadRoot
  <npmDownloadRoot>http://artifactory/npm/-/</npmDownloadRoot>
```
to
```
  <nodeDownloadRoot>http://nodejs.org/nodejs/dist/</nodeDownloadRoot
  <npmDownloadRoot>http://npm.org/npm/-/</npmDownloadRoot>
```
or similar.

What do you think?

eirslett:
Yes, that would be better.

geneticgrabbag:
I'll close this pull request and create a new one with an updated example then.  Thanks!

PR #237, opened 2015-06-19
Enhance example of different download roots.

geneticgrabbag:
Change the download roots in the installation section of the README to show two different repos.
Shows reasonable URLs for NodeJS (nodejs.org) and NPM (npm.org).


eirslett:
Thanks a lot! :-)

PR #242, opened 2015-07-05
Maven prereq >= 3.1.0, added slf4j-simple to resolve missing StaticLoggerBinder error

liebowitz:
Hi, nice plugin. The mojo appears to be compatible with the latest Maven version -- It doesn't appear necessary to restrict it to 3.1.0.  I also added a dependency on slf4j-simple to resolve the missing StaticLoggerBinder error.

### DIFF
--- a/frontend-maven-plugin/pom.xml
+++ b/frontend-maven-plugin/pom.xml
@@ -12,7 +12,7 @@
 
     <name>Maven Frontend Plugin</name>
     <prerequisites>
-        <maven>3.1.0</maven>
+        <maven>>=3.1.0</maven>
     </prerequisites>
 
     <description>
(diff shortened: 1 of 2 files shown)


abner:
+1 I was used to use the plugin with maven 3.3
I had to downgrade to older version of plugin because of this added restriction

eirslett:
@stephenc Will including slf4j-simple in the plugin still work, or will we hit the previous class loader issues?

frontend-maven-plugin/pom.xml
```diff
@@ -12,7 +12,7 @@
 
     <name>Maven Frontend Plugin</name>
     <prerequisites>
-        <maven>3.1.0</maven>
+        <maven>>=3.1.0</maven>
```
stephenc: that's not a valid range syntax. you probably mean `[3.1.0,)`

liebowitz: Thanks.  Probably best to revert this change.

Issue #235, opened 2015-06-17
-DskipTests should be able to skip more than just Karma tests.


KenPowers:
It would be great to be able to skip tests run from a `scripts` directive in `package.json` through `npm`.

PR #259, opened 2015-08-05
Allow any Mojo to be skipped during test phase.

KenPowers:
This addresses #235. Instead of only skipping Karma tests with `-DskipTests` the `AbstractFrontendMojo` now checks to see if `skipTests` is set and if it is running in the `test` phase. If these conditions are met then the execution is skipped. Each additional mojo is still skippable via their respective `skip` configurations.

### DIFF
--- a/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/AbstractFrontendMojo.java
+++ b/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/AbstractFrontendMojo.java
@@ -39,7 +39,15 @@ public abstract class AbstractFrontendMojo extends AbstractMojo {
    * Determines if this execution should be skipped.
    */
   private boolean skipTestPhase() {
-    return skipTests && execution.getLifecyclePhase().equals("test");
+    return skipTests && isTestingPhase();
+  }
+
+  /**
+   * Determines if the current execution is during a testing phase (e.g., "test" or "integration-test").
+   */
+  private boolean isTestingPhase() {
+    String phase = execution.getLifecyclePhase();
+    return phase.equals("test") || phase.equals("integration-test");
   }
 
   protected abstract void execute(FrontendPluginFactory factory) throws FrontendException;
--- a/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/AbstractFrontendMojo.java
+++ b/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/AbstractFrontendMojo.java
@@ -4,12 +4,24 @@ import com.github.eirslett.maven.plugins.frontend.lib.FrontendException;
 import com.github.eirslett.maven.plugins.frontend.lib.FrontendPluginFactory;
 import com.github.eirslett.maven.plugins.frontend.lib.TaskRunnerException;
 import org.apache.maven.plugin.AbstractMojo;
+import org.apache.maven.plugin.MojoExecution;
 import org.apache.maven.plugin.MojoFailureException;
+import org.apache.maven.plugins.annotations.Component;
 import org.apache.maven.plugins.annotations.Parameter;
+import org.slf4j.LoggerFactory;
 
 import java.io.File;
 
 public abstract class AbstractFrontendMojo extends AbstractMojo {
+
+  @Component
+  protected MojoExecution execution;
+
+  /**
+   * Whether you should skip while running in the test phase (default is false)
+   */
+  @Parameter(property = "skipTests", required = false, defaultValue = "false")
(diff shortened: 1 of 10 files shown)


frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/AbstractFrontendMojo.java
```diff
@@ -23,13 +35,23 @@
   @Parameter(defaultValue = "${basedir}", property = "installDirectory", required = false)
   protected File installDirectory;
 
+  /**
+   * Determines if this execution should be skipped.
+   */
+  private boolean skipTestPhase() {
+    return skipTests && execution.getLifecyclePhase().equals("test");
```
jollytoad: Would it also be worth adding a check for "integration-test" lifecycle phase too?

KenPowers:
@jollytoad How's that?

jollytoad:
Awesome, great work.

eirslett:
How would this work if you want to only skip the karma tests, but not skip the grunt build?

KenPowers:
@eirslett The idea here is to skip anything in the `test` phase when `skipTests` is set. Builds, to me, belong in another phase altogether. If you wanted a more complex setup involving skipping or not skipping various different things in `node` then perhaps profiles would be a better option.

I am by no means a Java / Maven expert. I am actually a frontend developer and my team has started using this plugin recently. They expressed that they wanted to be able to skip tests but we want to run tests through `npm`.

eirslett:
Ah, I see. The usual way skipTests works in Maven, is that you run `mvn clean install -DskipTests`, which will compile all Java sources, skip tests, and then package the classes to a jar/war file. Similarly, `-DskipTests` for frontend stuff would run gulp/grunt/whatever to compile assets, skip tests, and then put the generated assets in a `src/main/webapp/assets` directory or something like that.

The easiest case to solve your use-case is to use a profile `-PskipTests` which sets the `<skip>` parameter to true, and it should default to false.

Pseudocode:

```xml
<properties>
  <skipNpmTests>false</skipNpmTests>
</properties>
...
<execution>
  <id>npm test</id>
  <configuration>
    <skip>${skipNpmTests}</skip>
  </configuration>
</execution>
...
<profile>
  <id>skipTests</id>
  <properties>
    <skipNpmTests>true</skipNpmTests>
  </properties>
</profile>
```
```

FagnerMartinsBrack:
FWIW, I recently started [an OSS project](https://github.com/js-cookie/java-cookie) that uses this plugin to run integration tests in the front-end along with Selenium/Arquillian/QUnit/Node. The tests use bower to run the tests from another related [JavaScript project](https://github.com/js-cookie/js-cookie/blob/master/CONTRIBUTING.md#integration-with-server-side). To be able to create the relevant artifacts without having to run the whole integration test suite, I created a profile named `simple` that is run as `mvn install -P simple`.

That could be taken as an example of how to do it.

See the original commit: https://github.com/js-cookie/java-cookie/commit/46de889ef237e38fb119e675906b7aab1c14d74f.

jollytoad:
Having the standard -DskipTests also skip frontend based tests seems like reasonable and expected behaviour to me. Adding an extra profile specifically for this just feels like a hack.

eirslett:
Yes, -.DskipTests should skip frontend tests, but it should still compile assets. The main problem with this PR is that when you run -DskipTests it will also skip asset compilation. One would need a configuration parameter, something like <isTest> I suppose, to know whether an npm call is asset compilation or test runs?

eirslett:
Oh wait, maybe I missed something; this PR differentiates between them by looking at which phase they are executed?

jollytoad:
Yep, from my understanding, it skips only executions in the test phases, so asset compilation should be unaffected in the generate-resources phase.

KenPowers:
Yes, this only skips during testing phases hence my comment:

> Builds, to me, belong in another phase altogether.

KenPowers:
> One would need a configuration parameter, something like I suppose, to know whether an npm call is asset compilation or test runs?

That is actually something I considered, but I thought it would be cleaner to just detect the phase of the current execution.

KenPowers:
Any thoughts about reconsidering this? I'd prefer to avoid `<profile>`s if possible since `-DskipTests` is fairly standard.

ddumontatibm:
@eirslett It seems like your concerns have been addressed since you closed the pull request.  Can you please re-open and reconsider this?

ddumontatibm:
@eirslett any news about re-opening this pull?

brianc:
yeah we're feeling some pain on this too - massive java project w/ a front-end component...need the tests to execute in the test phase and be skipped regardless of the phase if `-DskipTests` is passed to maven.

eirslett:
Ok guys, I'll let this one in after all!

PR #260, opened 2015-08-07
Implemented reading properties from package.json

KnisterPeter:
This new mojo does reading properties from the package.json file and injects them into the maven build.
This is useful to give build artifacts the version number or name from the package.json.

### DIFF
--- a/frontend-maven-plugin/src/it/example project/pom.xml
+++ b/frontend-maven-plugin/src/it/example project/pom.xml
@@ -79,6 +79,13 @@
                         </configuration>
                     </execution>
 
+                    <execution>
+                        <id>read package.json</id>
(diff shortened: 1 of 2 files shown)


eirslett:
Is it possible that this feature can live in a separate maven plugin? package-json-maven-plugin? I realise it has its use cases, but I think it provides a slightly different feature than the node/npm integration this plugin gives you. (And I'm afraid of feature bloat)

KnisterPeter:
Okay. I'll move to a separate plugin. I thougt it would be nice to have the
package.json information available in the maven process.

Eirik Sletteberg <notifications@github.com> schrieb am So., 9. Aug. 2015
16:24:

PR #264, opened 2015-08-16
Suggestion: helper scripts are not really necessary

klieber:
Since you already have scripts for node and npm in the `/node` directory and npm symlinks in scripts for the other tools into `/node_modules/.bin` you really don't need the helper scripts.  All you need to do is add those two directories to your path.  So, just a suggestion to remove the helper scripts and include instructions for adding those directories to the path instead.


mriehema:
:+1:

eirslett:
Sorry again for being so late with all the PRs, but I'll close this one since I think it's better to have both options available. (Maybe only document the method of using PATH instead of helper scripts)

PR #267, opened 2015-08-24
Improved proxy support

bvklingeren:
Current proxy support is somewhat naive, retrieving only the first active proxy available in the Maven settings. This PR contains an improvement that retrieves all (active) proxies and uses the first matching based on the protocol used (so when downloading a file from http, the http proxy is used) for downloading node and npm.
When executing npm, both npm arguments --proxy and --https-proxy are set when corresponding proxies are specified.

### DIFF
--- a/frontend-maven-plugin/src/it/settings.xml
+++ b/frontend-maven-plugin/src/it/settings.xml
@@ -8,6 +8,24 @@
       <mirrorOf>*</mirrorOf>
     </mirror>
   </mirrors>
+    <proxies>
+        <proxy>
+            <id>http</id>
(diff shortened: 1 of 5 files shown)


eirslett:
LGTM, thanks!

bvklingeren:
No problem, thank you for developing/maintaining the plugin! Do you have an idea when the next release is scheduled?

eirslett:
I'm sorry, I don't know... I'll try to merge some PRs when I have time and then release. (I don't have too much time to maintain this project.) What you could do until then is to build the latest version of the plugin for yourself and use that.

PR #269, opened 2015-08-29
install directory should default to working directory

klieber:
The install directory will now default to the working directory if it hasn't been specified.  I've also updated the README file to mention the `installDirectory` parameter.


eirslett:
Thanks! :-)

Issue #99, opened 2014-10-20
Problem downloading node


FagnerMartinsBrack:
Hi @eirslett,

Sorry to bother with an issue, but I deployed recently a project using `frontend-maven-plugin` on Travis -CI and it is not being able to download Node.js. See [here](https://travis-ci.org/FagnerMartinsBrack/WebStories/builds/38548800) in the bottom of the console.

It says:

``` shell
[ERROR] Failed to execute goal com.github.eirslett:frontend-maven-plugin:0.0.16:install-node-and-npm (install node and npm) on project WebStories: Could not download Node.js: Could not download http://nodejs.org/dist/v0.10.32/node-v0.10.32-linux-x64.tar.gz: sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target -> [Help 1]
```

Any idea why this is happening? The local build on Windows works perfectly =/

PR #270, opened 2015-08-31
Default nodejs root to https

fiskr:
NodeJS sometimes doesn't permit http requests, which can cause build failures.
This has [happened before](https://github.com/eirslett/frontend-maven-plugin/issues/99) with negative consequences for others.

While it's possible to override using the `<documentRoot>` configuration, I suggest changing the default to use https instead of http.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ Node/npm will only be "installed" locally to your project. It will not be instal
     <configuration>
         <nodeVersion>v0.10.18</nodeVersion>
         <npmVersion>1.3.8</npmVersion>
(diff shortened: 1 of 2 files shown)


mriehema:
+1

fiskr:
@eirslett - how do you feel about merging this in?

eirslett:
Let's do it! Thanks for the PR!

kmila:
Waiting for README (_npmDownloadRoot_) update to http://registry.npmjs.org/npm/-/ (line 95 or so).

PR #283, opened 2015-09-11
This close #271

### LINKED ISSUE #271: Add the ability to disabling proxy settings for npm
We would like to proxy the download of npm and node, but not the npm install. Can we add a flag to disable passing the proxy settings to npm?

drautureau:
This PR honors the nonProxyHosts for npm commands and give the possibility to define the NPM registry for npm commands (useful if you proxy it using binary repositories such as Artifactory or Nexus)

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,13 +97,6 @@ You can also specify separate download roots for npm and node as they are now st
 </execution>
 ```
 
-You can also specify a proxy of the NPM registry (i.e. proxied through repository managers), npm will be download from it if npmDownloadRoot is not set.
-```xml
-<configuration>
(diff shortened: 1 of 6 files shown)


eirslett:
I think you can already do this with `.npmrc` or `package.json`? In which case it will also work if you run npm outside of Maven :-)

drautureau:
ah sorry, I am not used to deal with this tools at all!
I did not want to use the `~/.npmrc` file because npm commands do not use it in this plugin and I did not want to overwrite the user environment (could break other builds).

Please also notice that this PR honors the nonProxyHosts for npm commands, otherwise proxy parameters are added to the npm command line.

Will dig into `package.json` deeper!

thanks

PR #284, opened 2015-09-15
node v4.0.0 download path fix

### LINKED ISSUE #276: Can't download Node v4.0.0 because the URL has changed
This plugin breaks when trying to download Node v4.0.0. The url to download Node has changed. Previously it was:
http://nodejs.org/dist/v0.12.2/node.exe
http://nodejs.org/dist/v0.12.2/x64/node.exe

Now it is:
http://nodejs.org/dist/v4.0.0/win-x86/node.exe
http://nodejs.org/dist/v4.0.0/win-x64/node.exe

alalpod:
This closes  #276. For node version 4.0.0 has changed path


eirslett:
Thanks a lot! Could you please follow up with a PR that does `if nodeVersion > 4.0.0` or something like that? I suspect/hope they'll continue with the new scheme.

alalpod:
Ok, I'll try, but I'm not a Java developer, just need a little time to try it out

alalpod:
ok, sorry, will do in 10 minutes

alalpod:
@eirslett show I do new PR?

PR #285, opened 2015-09-15
node v4.0.0 download path fix

alalpod:
Fixed PR #284 and added comparation of the versions. For all versions >= v4.0.0 there will be new path.
I've tested it, but still not sure, I'm new to Java, so please check if it's alright.
Thanks.

### DIFF
--- a/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/Platform.java
+++ b/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/Platform.java
@@ -67,13 +67,13 @@ class Platform {
     public String getNodeDownloadFilename(String nodeVersion) {
         if(isWindows()) {
             if(architecture == Architecture.x64){
-                if (nodeVersion == "v4.0.0") {
+                if (nodeVersion.equals("v4.0.0")) {
                     return nodeVersion+"/win-x64/node.exe";    
                 } else {
                     return nodeVersion+"/x64/node.exe";
                 }
             } else {
-                if (nodeVersion == "v4.0.0") {
+                if (nodeVersion.equals("v4.0.0")) {
                     return nodeVersion+"/win-x86/node.exe";    
                 } else {
                     return nodeVersion + "/node.exe";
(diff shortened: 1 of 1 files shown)


frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/Platform.java
```diff
@@ -66,14 +66,15 @@ public String getLongNodeFilename(String nodeVersion) {
 
     public String getNodeDownloadFilename(String nodeVersion) {
         if(isWindows()) {
+			int versionCompare = nodeVersion.compareTo("v4.0.0");
             if(architecture == Architecture.x64){
-                if (nodeVersion == "v4.0.0") {
+				if (versionCompare >= 0) {
```
FagnerMartinsBrack: `"v4.0.0".equals( nodeVersion )` would suffice without diff verbosity.
It seems there are some spacing changes here.

eirslett:
compareTo used when comparing strings is about alphabetical order, so it's a < b, b < c and so on. I think the comparison needs to actually understand semver version numbers, which requires some sort of Java library that can do that for us...

frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/Platform.java
eirslett: The point was to use the new URL mapping for Node.js versions greater than 4, so that when 4.1.0 is released, the code would still work...

jollytoad:
I think that you could just assume that anything non v0.x.x will follow the new convention, without resorting to dissecting the version number, eg:

    if (nodeVersion.startsWith("v0.")) {
       ... old path ...
    } else {
       ... new path ...
    }

eirslett:
> I think that you could just assume that anything non v0.x.x will follow the new convention, without resorting to dissecting the version number, eg:

Yes, thanks for the suggestion, that's much easier. Let's do that!

frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/Platform.java
FagnerMartinsBrack: I thought this was just fixing the wrong comparison [of the last PR](https://github.com/eirslett/frontend-maven-plugin/commit/fa6430d9bbf15b4d3a99a3ca63beda900b54c03d). Sorry 'bout that.

alalpod:
Thanks @jollytoad, it's a realy great idea.

eirslett:
Thanks, let's try this :-)

PR #300, opened 2015-10-13
add grunth support for ES6 features

jackTheRipper:
Hi,

I added grunth to your plugin in order to be able to solve the problem we have with Grunt-cli and ES6 (https://github.com/gruntjs/grunt-cli/issues/58).

If you have any comments or remarks ask me !

### DIFF
--- a/frontend-maven-plugin/src/it/example project/package.json
+++ b/frontend-maven-plugin/src/it/example project/package.json
@@ -5,6 +5,7 @@
     "bower": "~1.3.12",
     "grunt": "~0.4.1",
     "grunt-cli": "~0.1.9",
+    "grunth-cli": "^1.3.1",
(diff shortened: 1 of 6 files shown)


jackTheRipper:
@eirslett would it be possible to do something with my PR? I have some requirements with one of my project for the ES6 launcher

eirslett:
I think the best way to achieve this is to add a script to your package.json file and then run grunt through the npm mojo instead, with `<arguments>run-script grunt</arguments>`.
Since there are so many new build tools and potential additional mojos, we might refactor this plugin, remove all build-tool-specific mojos, and delegate all responsibility for task execution to the npm mojo.

Does that work for you?

jackTheRipper:
Ok let's do so. I cancel my pull-request

PR #303, opened 2015-10-22
Example shouldn't link to the National Association of Pastoral Musicians

cleiter:


eirslett:
That's a nice easter egg, @geneticgrabbag - it was finally discovered :smirk:

geneticgrabbag:
Nice! :)

PR #316, opened 2015-11-28
Fixed unhandled NullReferenceException in specific case

jelical:
There is NullReferenceException when executing plugin goal directly or as 'default-cli' in 'noTests' mode  because in such case lifecycle phase is undefined.


tscheinecker:
:+1: currently it is quite annoying to have to run all the tests every time I want to execute my frontend-build due to this fact

eirslett:
LGTM :+1:

PR #330, opened 2016-01-05
add support for caching downloaded files

klieber:
I've been using the `installDirectory` parameter to avoid downloading node and npm everytime a build runs on my CI server by configuring all my builds to use a common install directory that is outside of the working directory.  This has been working fine but now I want to upgrade to newer versions of node and npm.  If I upgrade then the build will run and wipe out the old version from the install directory and install the new version which is great except for the fact that I still have builds of the older code running as well.  So basically it just keeps flipping back and forth between versions.  So far two builds with different versions haven't run at the same time but I'm sure that will happen soon and who knows what the result will be.

This pull requests adds a new optional parameter `cacheDirectory` which if specified will be used to store the downloaded files by name and version.  This way I can have multiple builds using the same cache directory so they don't need to download them everytime and if the version changes they don't cause each other problems.

I also cleaned up the logging a little bit and updated the example project to use newer versions of everything.

### DIFF
--- a/frontend-maven-plugin/src/it/example project/package.json
+++ b/frontend-maven-plugin/src/it/example project/package.json
@@ -2,24 +2,26 @@
   "name": "example",
   "version": "0.0.1",
   "dependencies": {
-    "bower": "~1.3.12",
-    "grunt": "~0.4.1",
-    "grunt-cli": "~0.1.9",
-    "grunt-contrib-jshint": "~0.6.4",
-    "grunt-karma": "~0.7.0",
+    "bower": "~1.7.2",
+    "grunt": "~0.4.5",
+    "grunt-cli": "~0.1.13",
+    "grunt-contrib-jshint": "~0.11.3",
+    "grunt-karma": "~0.12.1",
     "gulp": "^3.9.0",
-    "jspm": "~0.15.6",
-    "karma": "~0.12.0",
-    "karma-jasmine": "~0.1.5",
-    "karma-phantomjs-launcher": "~0.1.1"
+    "jspm": "~0.16.19",
+    "karma": "~0.13.18",
+    "karma-jasmine": "~0.3.6",
+    "karma-phantomjs-launcher": "~0.2.3",
+    "jasmine-core": "^2.4.1",
+    "phantomjs": "^1.9.19"
   },
   "jspm": {
     "dependencies": {
       "jquery": "github:jquery/jquery@^2.1.3"
     },
     "devDependencies": {
-      "traceur": "github:jmcriffey/bower-traceur@0.0.88",
-      "traceur-runtime": "github:jmcriffey/bower-traceur-runtime@0.0.88"
+      "traceur": "github:jmcriffey/bower-traceur@0.0.95",
+      "traceur-runtime": "github:jmcriffey/bower-traceur-runtime@0.0.95"
     }
   },
   "scripts": {
--- a/frontend-maven-plugin/src/it/example project/pom.xml
+++ b/frontend-maven-plugin/src/it/example project/pom.xml
@@ -23,8 +23,8 @@
                             <goal>install-node-and-npm</goal>
                         </goals>
                         <configuration>
-                            <nodeVersion>v0.12.2</nodeVersion>
-                            <npmVersion>2.7.6</npmVersion>
+                            <nodeVersion>v5.3.0</nodeVersion>
+                            <npmVersion>3.3.12</npmVersion>
                         </configuration>
                     </execution>
 
--- a/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/AbstractFrontendMojo.java
+++ b/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/AbstractFrontendMojo.java
@@ -8,6 +8,8 @@ import org.apache.maven.plugin.MojoExecution;
 import org.apache.maven.plugin.MojoFailureException;
 import org.apache.maven.plugins.annotations.Component;
 import org.apache.maven.plugins.annotations.Parameter;
+import org.apache.maven.project.MavenProject;
+import org.eclipse.aether.RepositorySystemSession;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
@@ -43,6 +45,18 @@ public abstract class AbstractFrontendMojo extends AbstractMojo {
    @Parameter
    protected Map<String, String> environmentVariables;
 
+  @Parameter(
+      defaultValue = "${project}",
+      readonly = true
+  )
+  private MavenProject project;
+
+  @Parameter(
+      defaultValue = "${repositorySystemSession}",
+      readonly = true
+  )
+  private RepositorySystemSession repositorySystemSession;
+
   /**
    * Determines if this execution should be skipped.
    */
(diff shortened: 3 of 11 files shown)


mriehema:
I checked the commit. Awesome, I like it. :+1:

wimdeblauwe:
+1 for this!

eirslett:
My main concern is that the plugin is gaining a lot of weight and complexity. We now have workingDirectory, installDirectory, downloadRoot, nodeNownloadRoot, npmDownloadRoot and cacheDirectory - it was all supposed to be simple, but somehow ended up being very confusing... is there a reason you cannot go with the default settings of the plugin? Which is to download one node instance for every build, and cache it in the "node" folder so that it's only downloaded the first time you run the build? (I guess that's what you're after, in the end?)
I also think some people have used downloadRoot with a `file://` protocol, to use local caches for the node executables.

klieber:
I don't understand your point on the complexity issue but I think you have to weigh the benefits against it.  The code really isn't that much more complex and as far as understanding it's use I think that is more of a documentation issue.

The reason the default settings don't work is because on our build servers we use a fresh workspace for every build. This is pretty typical of most build systems (like Travis CI for instance). The problem with the custom downloadRoot is that if I go that approach I have to maintain a mirror of the npm and node repositories so that the build doesn't break if someone upgrades the node/npm version. We have 20+ developers on our team so this is very likely. The other problem is that I have multiple build server slaves so I would have to maintain this mirror on each of them.

I support two open source maven plugins so I really do get where you are coming from on want to stop the config parameter madness but I do really think this one is needed. (I actually think most people would be fine with getting rid of the install directory parameter if we had this one instead.)

klieber:
There is one other way to handle this if your main concern is adding the new config parameter.  I implemented caching in my [phantomjs-maven-plugin](https://github.com/klieber/phantomjs-maven-plugin) using the local maven repository via it's API.  This means that the artifact is always cached but there is no need for a user to configure it because it's just cached in their local maven repository like all other maven artifacts.  This is a really nice way to go about it but does require a bit more work on the coding side which is why I avoided it for now.  It really wouldn't be that bad though so if you'd rather go that way I'd be happy to implement it when I get some time.

eirslett:
Yes, putting it in .m2 is also a way to go, which I also discussed with the author of the similar gradle plugin. Then, it would also be possible to distribute Node packaged in jars (like the webjars project) and use Maven's in-built mechanism for downloading them.

mriehema:
Wow, this would be the best solution.

klieber:
I've updated the pull request so that now the maven local repository is used for the cache.  No new parameters are necessary.

eirslett:
I prefer to keep the mojos as small as possible, and move this login into frontend-plugin-core, so that potentially other build systems can use the same code. Would that be possible?
I'm thinking of something like creating the .m2 folder if it doesn't exist, and then _not_ using the default Maven repository layout, but rather something like .m2/frontend-maven-plugin/xxx. That will help for forwards-compatibility if we ever get to the point where node/npm is directly fetched as "proper" Maven artifacts.

rnc:
@eirslett  What happens if the user is not using the default .m2 but another location set via `-Dmaven.repo.local` (or via a settings.xml) ?

eirslett:
We'd have to ignore that case and put it in ~/.m2 anyways. Or maybe the m2 repository logic is loosely coupled from Maven itself? So it's a library people can use? What does sbt and gradle do? They too use the m2 directory, I think...

klieber:
@eirslett, I'm not sure what you mean by "move this login into frontend-plugin-core".  What login?

The way I designed this was to keep the frontend-plugin-core library free of dependencies on maven.  That is why there is simply just a `CacheResolver` interface in frontend-plugin-core and then whomever is using it can implement it for their purposes.  So, like you said if someone what's to use it in another build system they just need to implement it for their build system.  If you wanted the `RepositoryCacheResolver` moved into frontend-maven-core then we would need dependencies on maven and then those not using maven would be pulling down unnecessary dependencies.

On the topic of directly fetching node/npm as "proper" Maven artifacts, is this something you really want to do?  You would then be responsible for uploading all versions of node/npm up to central and node typically has 2-3 releases a month.  I currently do this for phantomjs-maven-plugin but they have a much slower release cycle so it's more feasible for me (only two releases in the last year).  If you did want to move along this route that also introduces more dependencies on maven.  You will need to inject in the `RepositorySystem` and a list of `RemoteRepository` instances and then the code is basically this:

```java
RepositorySystem repositorySystem = // .. from maven
RepositorySystemSession repositorySystemSession = //.. from maven
List<RemoteRepositories> remoteRepositories = //.. from maven

Artifact artifact = new DefaultArtifact("com.github.eirslett", "node", "x64", "tar.gz", "4.2.6");

ArtifactRequest request = new ArtifactRequest();
request.setArtifact(artifact);
request.setRepositories(remoteRepositories);

ArtifactResult result = repositorySystem.resolveArtifact(repositorySystemSession, request);

File file = result.getArtifact().getFile();
```

The library maven uses to do this is called [Eclipse Aether](http://www.eclipse.org/aether/).  I believe Gradle uses aether internally for supporting maven repositories but it is not intended to be used directly because it has it's own dependency resolution system.  So again, to implement resolving dependencies from maven I would just put an interface in the frontend-plugin-core and then implement that in the frontend-maven-plugin and others using frontend-plugin-core with another system like Gradle can implement it using Gradle's dependency resolution system.

eirslett:
I'm sorry, not sure why I wrote "login". I guess I meant "code" or "functionality"...
Ok let's keep this in Maven for now, and let people inject a cache resolver.
What happens if no resolver is specified? I guess it will fallback to downloading directly, no caching?

It also looks like the build failed :-(

MTyson:
When you said "login" there I read "logic" :P

klieber:
@eirslett, build should be fine now.  I just forgot to add something to the commit.  At the moment a cache resolver implementation needs to be provided.  I could make it so that if one isn't provided then a default implementation is used which just caches in the node directory if you think that is necessary.

eirslett:
I think CacheProvider should be opt-in. So if you don't need caching, you won't even need to provide a CacheProvider (e.g. another project using frontend-plugin-core).

klieber:
Updated so that there is a default cache provider used if one is not provided.  So if you don't want to use one you don't have to.  The default will just stick the downloaded files in the install directory just like it does today.

mriehema:
I followed your discussion and I like this feature. Some of our developers are traveling with no network connection and they building our workspace with {{mvn clean install --offline}}. So we need a way to cache all node_modules too. I took a look into the changes, but didn't find any documentation or examples of what I have to do, to enable caching (in the .m2 repository) by default. After merging and releasing this changes, I would adapt this asap and give some feedback.

klieber:
@mriehema this change is only about caching `node` and `npm` and nothing to do with the modules.  I believe npm should already be caching modules for you.  Try running `npm cache ls` and you will see everything it has cached.

klieber:
@eirslett is there anything else you are waiting on me to change here?  I'm using a custom build of this at work and it would be great to get back on the official build.

eirslett:
I'm sorry, it's just me not paying enough attention... I had a proper look at it now, it looks good. Thanks!

wimdeblauwe:
Any ETA on a release with this?

eirslett:
It's released now.

wimdeblauwe:
Ah, ok. Are there release notes somewhere of the various releases? And documentation on the configuration parameters?

eirslett:
No, unfortunately not. For now, the git log is the release/changelog, and the integration tests serve as documentation. PRs for the most important things to document in README.md would be nice, other things can be documented in the wiki.

Issue #343, opened 2016-01-21
npm WARN generates a Maven ERROR


mriehema:
In my npm install phase some node_modules generate some warnings about deprecated versions (like grunt). But in the Maven log, this is an ERROR instead of a WARNING. We're using Jenkins as our build tool and these errors cause wrong log analysis.

```
[INFO] --- frontend-maven-plugin:0.0.27:npm (install node, npm and dependencies) @ theme ---
[INFO] Found proxies: []
[INFO] Running 'npm install --color=false' in C:\theme
[ERROR] npm WARN deprecated lodash@0.9.2: lodash@<2.0.0 is no longer maintained. Upgrade to lodash@^3.0.0
[ERROR] npm WARN deprecated lodash@1.0.2: lodash@<2.0.0 is no longer maintained. Upgrade to lodash@^3.0.0
[ERROR] npm WARN deprecated npmconf@2.1.2: this package has been reintegrated into npm and is now out of date with respect to npm
[ERROR] npm WARN prefer global node-gyp@3.2.1 should be installed with -g
```

Changing the logging from ERROR to WARNING would be nice.

mriehema:
Awesome! Thank you. :+1:

Q1nt:
+1 
having problems with bower warn messages

breun:
+1

ysangkok:
Thanks for the fix. Will a release be made now for this fix or should I use Jitpack to use this from master?

eirslett:
I didn't know about Jitpack, it looks nice! Go with that for now :)

breun:
It would be nice if it would be possible to see in what release a particular issue was (or will be) fixed. I can't find that information on an issue page and the releases at https://github.com/eirslett/frontend-maven-plugin/releases also don't have any changelogs, so I never know what changes I'm getting when I update to a new version.

eirslett:
That's right, there's no changelog... you'd have to look at the commit log.

breun:
Maybe attach issues to version milestones (vX.Y.Z or something)? That way it's also easy to generate a link to a list with all issues associated with a specific version milestone, which would essentially be a changelog for a release.

mriehema:
Hey @eirslett, I will create a pull request the next days with a CHANGELOG.md for the last versions. Afaik, it will automatically be attached to the [releases ](https://github.com/eirslett/frontend-maven-plugin/releases) page. Shouldn't be to much work to keep it updated.
[...]

PR #345, opened 2016-01-21
Fixes #41 #343 rewrites ProcessExecutor using commons-exec

### LINKED ISSUE #41: Node keeps running when killing maven
When running a Grunt watch task and hitting "Ctrl-C" to stop the process. Node.exe keeps running in the background.

I haven't found a way to kill node.exe other than going manually in the task manager...

We are using a watch task to compile sass files in the background.

tjuerge:
By using `commons-exec` instead of  plain `ProcessBuilder` we can support killing the Node process if the Maven build is stopped (via `ShutdownHookProcessDestroyer`). An optional timeout (which leverages `ExecuteWatchdog`) can be specified (not exposed to the task runners yet).

With `commons-exec` we get quote handling and variable substitution in arguments for free (via `CommandLine.addArgument()`).

A workaround for #343 is implemented as well.


eirslett:
Looks like the build failed though...

tjuerge:
It seems that the quote handling of arguments (which ``commons-exec`` enables by default) is the culprit for the build failure on unix. Disabling quote-handling fixes the build :)

mriehema:
@tjuerge Could you resolve the conflicts for this pull request? Otherwise I would create a different pull request just for #343.

mriehema:
Please close this PR, see #561 instead.

Issue #180, opened 2015-03-23
color on console


dokuboyejo:
How can the default argument option --color=false be changed.
I will like to have a colored output log

eirslett:
In the early stages of this plugin, it was decided to remove all colors from the log, because it interfered with build servers that couldn't show colors and instead showed strange artifacts/symbols.
It might make sense to eigher (1) re-add colors as the default option or (2) add a <colors>true/false</colors> option. The problem is that each runner (grunt, gulp, bower...) has its own way of specifying colors or not, with different names for the flags.
I'd be happy to merge a PR that adds a <colors> property! But for now I'll close this issue.

jgangemi:
this is still open in #122. 

i haven't looked at what's going on in the code as of late, but why not just remove whatever forces the colors to be off and let the users specify to turn them on in the args?

eirslett:
Initially, it was turned off because many logs were showing strange characters instead of colors (the ones without color support).

Issue #122, opened 2014-11-28
allow running grunt w/ colors to be optional


jgangemi:
i realize i'm rather late to this party but i like having the color output in my maven logs and would like to see this available as an option instead of being outright disabled.

my `TERM` is set to `xterm-256color` which may explain why everything renders correctly on the screen for me.

eirslett:
I'll close this one and wait for a PR, if it's still needed!

PR #347, opened 2016-01-25
Enables colors for all runners by default.

MadaraUchiha:
After reading #180 and #122, I've seen that you're waiting for a PR.

People with build systems that bork on colors can disable the specific platform's colors via the arguments passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ __Proxy settings:__ If you need to download Node/npm through a proxy: [configure
 
 ### Running npm
 All npm modules will be installed in the `node_modules` folder in your working directory.
-By default, no colors will be shown in the log.
+By default, colors will be shown in the log.
 ```xml
 <execution>
     <id>npm install</id>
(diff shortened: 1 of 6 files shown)


eirslett:
Thanks! Let's remove the parameter from the abstract class as well?

MadaraUchiha:
I haven't found any further instance of "color" in the entire project. I did miss the README though, I'll be pushing an update to that soon.

eirslett:
It's called "additionalArguments" in NodeTaskExecutor. Is it used somewhere else?

MadaraUchiha:
Yes, It's still used by DefaultBowerRunner's constructor (with `buildArguments(proxyConfig)`) and with DefaultWebPack runner, with an empty list (I'll just go ahead and remove that). 

The proxyConfig are passed from BowerMojo in line 41, where it gets generated with
```java
ProxyConfig proxyConfig = MojoUtils.getProxyConfig(session, decrypter);
```

I'm not really into Maven plugin development, so I'm not entirely sure what this method does. I'm fairly sure it's not related to output coloring. (I'm also not sure if this proxy configuration needs to use the additional arguments parameter, I don't know what's being passed there).

eirslett:
Ok, let's leave it for now... I'm not quite sure if the proxy stuff uses it or not.
Thanks for the PR!

PR #353, opened 2016-01-29
Use proxies more correctly. Fixes #322

### LINKED ISSUE #322: ProxyConfig doesn't allow http proxy for https host
It is not possible to build a project from behind a firewall if they use this plugin and override the default node registry url to an https url and your local proxy runs on http. 

The ProxyConfig (line 32) class checks that your configured proxy matches the URI scheme of the request URI this is not correct behaviour.

As a small fix I propose it be changed to:
```
 for (Proxy proxy : proxies) {
            if (!proxy.isNonProxyHost(uri.getHost())) {
                return proxy;
            }
        }
```

I don't believe this class should exist, the proxy should just be taken care of by the client.

I tried to patch it myself but can't build the project due to it not even finding any proxies when building, unsure why as I can use the released version 0.0.26 from maven central in other projects and it works fine.

Tim

timja:
ProxyConfig no longer checks the host scheme and compares it to the
configured proxy, it just uses the maven proxy now.

### DIFF
--- a/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ProxyConfig.java
+++ b/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ProxyConfig.java
@@ -30,11 +30,11 @@ public class ProxyConfig {
         }
         final URI uri = URI.create(requestUrl);
         for (Proxy proxy : proxies) {
-            if (proxy.protocol.equals(uri.getScheme()) && !proxy.isNonProxyHost(uri.getHost())) {
+            if (!proxy.isNonProxyHost(uri.getHost())) {
                 return proxy;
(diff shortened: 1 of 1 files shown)


bdellegrazie:
Hi @eirslett,
Is there any chance of getting this merged and a new release made?
The settings.xml 'protocol' part of the proxy configuration represents the proxy's protocol, not the client making the request. Maven doesn't support separate http / https proxies as only one proxy configuration can be active at a time.

I realise there are other PRs also addressing separate proxy issues but this fix corrects the *current* behaviour with maven proxies.
At present if you have a http proxy configured and the target is https it will go direct rather than via the proxy.

timja:
@eirslett any change this can be merged?

Tim

eirslett:
Ok, thanks for the PR!

PR #354, opened 2016-01-29
Updated README with example for maven 2

fidel-karsto:
With Version 0.0.17 org.codehaus.plexus.util.Scanner was introduced in MojoUtils.java, which is a dependency of maven 3 but not maven 2. 
The turn to make the plugin maven 3 only was done in version 0.0.23.
So we need to make the difference clear for people still depending on maven 2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,26 +39,7 @@ Include the plugin as a dependency in your Maven project.
     </plugin>
 ...
 ```
-## Maven 2
-```xml
-<plugins>
-    <plugin>
-        <groupId>com.github.eirslett</groupId>
-        <artifactId>frontend-maven-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


eirslett:
This is a rare edge case (people still using Maven 2), could you put it in the wiki instead? And add a simple a href link from the README to the wiki?

fidel-karsto:
Good to know that people maintaining enterprise software are working on rare edge cases! 😉
But you are right, the wiki is a good place for the hint as well. I will revert the readme and put it in the wiki.

fidel-karsto:
@eirslett I moved the maven 2 example to the wiki and linked to it from the README.md.

eirslett:
Thanks a lot! :-)

Issue #356, opened 2016-01-29
npmRegistryURL parameter does not work like i expect it to


jgangemi:
it appears the newly added `npmRegistryURL` needs to be set via system property and is not read directly out of the pom configuration and passed into `DefaultNpmRunner`.

can this be changed? i'm happy to submit a PR

eirslett:
I'll, close this as a duplicate of #357 .

PR #357, opened 2016-01-29
allow 'npmRegistryURL' to be set as part of the pom configuration

jgangemi:
this PR is for #356, it allows the `npmRegistryURL` to be configured in the pom as well as being specified via a `-D` option.


eirslett:
I think it's better to add npmRegistryURL directly in `<arguments>` or in `.npmrc`. Isn't that possible for you?

jgangemi:
not really - i have a couple of builds that share some parts of the configuration (such as the registry) but not others when it comes to arguments. the `.npmrc` file doesn't really help either b/c the build is meant to be portable and there's no where it can be shared that allows that.

i saw support for this got added, but it was only possible using a `-D` flag and i wanted to set it in the pom configuration.

eirslett:
Ok, let's get this merged then! Thanks!

PR #363, opened 2016-02-13
Use proxy for registry Url if given

pwielgolaski:
Thanks to that, we can eliminate usage of proxy that was defined as non-proxy for registry

It could be helpful if you have corporate NPM registry and do not want to use proxy for accessing to it.


eirslett:
The preferred way to solve this is by configuring a proxy in .npmrc (or package.json?) or passing it explicitly as a commond line option in `<arguments>`. That way, frontend developers will be able to run the build independent of maven. (And also, less magic involved.)
Is there a reason you cannot do that?

pwielgolaski:
The problem that I have is that frontend plugin does not respect non-proxy setting in maven.
When I have NPM registry in corporate network I dont need proxy, but I need it for maven if I want to grab something behind proxy.
So I can say that in this case I would like to disable proxy or obey non-proxy configuration.

LukeSkyw:
I know this is a one year old issue, but it is bitting us now.
I have a proxy configured in maven, but also a nonProxyHost option for internal servers. Our npm registry is on the corporate network, so we would expect either the maven-frontend-plugin not to pass the proxy to npm or to pass a no-proxy option too for npm to sort it out.

pwielgolaski:
Solved by https://github.com/eirslett/frontend-maven-plugin/pull/387

PR #368, opened 2016-02-23
Implement getAbsoluteTaskLocation to work similar to the standard NPM

Antibrumm:
task locator: Traverse up the parent directory until we find the
taskLocation. (Does not go global finally, if not found)

This supports multi module projects to consolidate node_modules on the
parent module:

parent -> frontend plugin install
module -> frontend plugin grunt / karma etc.

### DIFF
--- a/frontend-maven-plugin/pom.xml
+++ b/frontend-maven-plugin/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <artifactId>frontend-plugins</artifactId>
         <groupId>com.github.eirslett</groupId>
-        <version>0.0.28-SNAPSHOT</version>
+        <version>0.0.28-UP</version>
     </parent>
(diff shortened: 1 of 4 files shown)


Antibrumm:
Any interest in this?

eirslett:
Isn't this easier to solve by building an absolute url, based on Maven's variables like ${project.basedir}?

Antibrumm:
Initially I have tried to use a mix of install and working directories and couldn't get it running.  
NPM is going up the directory structure to resolve dependencies and task scripts, but for the initial call to "grunt" or "karma" this resolution is not done by the task runners. Is there a way to tell the plugin where the task scripts are located?

Antibrumm:
Closing this one as I think I found a better way

PR #373, opened 2016-02-26
Enable SSL client certificate authentication for node download url

benediktarnold:
Adds useSystemProperties to the httpclient builder in order to respect -Djavax.net.ssl.* properties. This would allow SSL client certificate authentication and custom truststores with self signed certificates to be used when downloading node and npm from a mirror that requires authentication.

A litle bit of background information:
We want to use this plugin in builds on our corporate jenkins. This jenkins has no direkt internet connection nor a connection via proxy. Instead we want to use a site repository on our nexus that mirrors nodejs.org/dist. Unfortunately the nexus requires client certificate authentication ;-)


eirslett:
LGTM! Thanks!

benediktarnold:
Great. Thanks for the quick merge

PR #374, opened 2016-02-29
Working with Maven 3.2.x and IDE console logging

beccari:
Everything works fine in my project, except we are using maven 3.2.5, so I had to remove the prerequisite tag. Probably is better to use *maven-enforcer-plugin* if the version constraint is really needed.

Also, I just added the  *slf4j-simple* dependency in order to get the logging working properly on my IntelliJ IDEA

```
[INFO] --- frontend-maven-plugin:0.0.29-SNAPSHOT:install-node-and-npm (install node and npm) @ example ---
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

### DIFF
--- a/frontend-maven-plugin/pom.xml
+++ b/frontend-maven-plugin/pom.xml
@@ -11,9 +11,6 @@
     <packaging>maven-plugin</packaging>
 
     <name>Maven Frontend Plugin</name>
-    <prerequisites>
-        <maven>3.1.0</maven>
-    </prerequisites>
 
     <description>
         This Maven plugin lets you install Node/NPM locally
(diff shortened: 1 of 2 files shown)


eirslett:
I suspect your project is being built by IntelliJ's own bundled, old version of Maven. I think it should work fine with Maven 3.2.5.

eirslett:
Actually, this project used to bundle slf4j-simple, but recent versions of Maven actually use slf4j, so we had to remove it as a dependency, to prevent clashing with Maven's version.

beccari:
You are right @eirslett ! I missed that! My intelliJ was using its internal bundled version. I will step down my request!
Tks a lot!

Issue #384, opened 2016-03-16
Allow proxy settings to be ignored


matthewhubblerose:
# My Setup

Build running on Jenkins behind a corporate firewall - all external HTTP traffic has to go through a proxy server. We have an internal Nexus repository - all traffic for that cannot go through the proxy server.

For that reason, we have something like this in our settings.xml:
```
<proxy>
  <active>true</active>
  <protocol>http</protocol>
  <host>proxy.company.com</host>
  <port>80</port>
  <nonProxyHosts>maven.company.com|*.other.com</nonProxyHosts>
</proxy>
```

The `nonProxyHosts` part is used to stop traffic for hosts matching the expressions being proxied.

# Problem

As far as I can tell (by grepping though the source code), non proxy hosts are not supported by npm - proxies are supported though. So when I build my maven project with something like this in the pom.xml:

```
<plugin>
    <groupId>com.github.eirslett</groupId>
    <artifactId>frontend-maven-plugin</artifactId>
    <version>0.0.22</version>
    <executions>
        <execution>
            <id>install node and npm</id>
            <goals>
                <goal>install-node-and-npm</goal>
            </goals>
        </execution>
        <execution>
            <id>npm install</id>
            <goals>
                <goal>npm</goal>
            </goals>
        </execution>
    </executions>
    <configuration>
        <nodeVersion>v5.7.1</nodeVersion>
        <npmVersion>3.6.0</npmVersion>
    </configuration>
</plugin>
```

 - install-node-and-npm succeeds as the requests are correctly proxied
 - npm fails as the traffic is incorrectly proxied

Specifically, the command run during npm is like this:
```
Running 'npm install --color=false --proxy=http://proxy.company.com:80' in /path/to/project
```

This means the requests to the Nexus repository are sent through the proxy server and fail.

# Temporary Workaround

It turns out there is a [--no-proxy command line option](https://github.com/npm/npm/commit/fec4c967ee235030bf31393e8605e9e2811f4a39) supported by npm that over-ride the use of a proxy if it has already been specified. So, running:
```
npm install --color=false --proxy=http://proxy.company.com:80 --no-proxy
```
means all of the Nexus HTTP traffic is not proxied. When you try to use the `--no-proxy` argument from the plugin configuration, like this:

```
<execution>
    <id>npm install</id>
    <goals>
        <goal>npm</goal>
    </goals>
    <configuration>
        <arguments>install --no-proxy</arguments>
    </configuration>
</execution>
```

[...]

PR #385, opened 2016-03-17
Fixes issue #384 by adding an noProxy configuration parameter npm execution

matthewhubblerose:
Fixes issue #384.

Also moved some logging around to make things less confusing.

### DIFF
--- a/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/InputStreamHandler.java
+++ b/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/InputStreamHandler.java
@@ -53,7 +53,12 @@ final class InputStreamHandler extends Thread {
         return new LogLevelAgnosticLogger() {
             @Override
             public void log(String value) {
-                logger.error(value);
+                // fix #343
+                if (value.startsWith("npm WARN ")) {
+                    logger.warn(value);
+                } else {
+                    logger.error(value);
+                }
             }
         };
     }


eirslett:
Maybe rename "noProxy" to "npmInheritsProxyConfigFromMaven", which defaults to true? It's a much longer property name, but also more descriptive. npm could still be configured with a proxy, even if this setting is configured.

matthewhubble:
Agreed - that's a much better name.

eirslett:
Looks like the npm dependencies in the example project/integration tests are out of date, I guess that's why the build is failing :-(

matthewhubble:
Looks like a problem with phantomjs. Wierd - as the tests pass a couple of hours ago with my first commit

matthewhubble:
```
[INFO] [ERROR] Error requesting archive.
[INFO] [ERROR] Status: 429
[INFO] [ERROR] Request options: {
[INFO] [ERROR]   "uri": "https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-1.9.8-macosx.zip",
```

matthewhubble:
Is it possible to re-run the checks? Perhaps it's was some temporary error - I managed to download from that URL no problem

matthewhubble:
It's a problem with the phantom-js download from:
https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-1.9.8-windows.zip

When I hit that URL, I get a response with this text:
Rate limit for this resource has been exceeded

Status code is 429 = Too Many Requests.

eirslett:
Looks like it worked now!

frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/NpmMojo.java
```diff
@@ -12,6 +12,7 @@
 import org.sonatype.plexus.build.incremental.BuildContext;
 
 import java.io.File;
+import java.util.ArrayList;
```
eirslett: Where is this used?

eirslett:
Looks good! It might also be worthwhile to add a short notice in README.md about proxies, and then link to the wiki with a longer article that explains this functionaliry?

matthewhubble:
I've added the documentation to the README. I think it explains everything there, I don't think the Wiki will also need updating.

frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/NpmMojo.java
matthewhubble: It was being used to create an empty list. I've now changed it to use `Collections.emptyList` to be more consistent with the rest of the code.

eirslett:
Thanks a lot! Could you please squash the commits into one? And then I'll merge it :-)

PR #408, opened 2016-04-28
Make using the maven caching optional

rnc:
While #330 is very useful, in certain constrained build environments caching files in the .m2 leads to verification errors when matching to deployed files. 

It would be useful to allow this new functionality to be optional.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,12 @@ You can also specify separate download roots for npm and node as they are now st
 </plugin>
 ```
 
+By default node/npm will be cached to to the Maven local repository. To disable this and use the `workingDirectory/cache` add
+
+    <configuration>
+        <useMavenCache>false</useMavenCache>
+
+
 ### Proxy settings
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -139,6 +139,12 @@ You can also specify separate download roots for npm and node as they are now st
 </plugin>
 ```
 
+By default node/npm will be cached to to the Maven local repository. To disable this and use the `workingDirectory/cache` add
```
eirslett: What do you mean workingDirectory/cache? Will a cache directory appear there by default? (I haven't used the cache functionality myself)

rnc: I believe by default it now downloads the npm (etc) tar.gz to e.g. .m2 directory instead of a temporary directory. The working directory is where you are running the build from. I guess we could make this more configurable? But I just wanted to the option to make it .m2 or something else :-)

rnc:
@eirslett : Any update on this please?

PR #420, opened 2016-05-16
Correcting link to Node.js download root in documenation

pdcamilleri:


eirslett:
Thanks!
Still, I'm concerned whether the documentation is misleading... people aren't supposed to configure these values unless they want to set them to something other than default. For example "http://nodejs.mymirror.example.com/dist/",

Issue #433, opened 2016-06-06
Setting NODE_ENV when running an npm goal


sshahar1:
I'm using the plugin to build angularJS code, with npm and webpack.
It works fine, but I need to set NODE_ENV, so some variables would be read from environments.json

I can't figure out if it is possible and how, so I am hoping for your help.

Thanks in advance
Sigal

adumat:
:+1: i need this!

sshahar1:
Working for webpack. I'm keeping this open for @sonirahul

adumat:
@sonirahul for me is working, but i'm invoking gulp using npm script from package.json

MatellioRahul:
@adumat can you please tell me how are you passing environment variables like NODE_ENV?

adumat:
@MatellioRahul with this definition on pom:
```
            <execution>
              <id>build</id>
              <phase>generate-resources</phase>
              <goals>
                <goal>npm</goal>
              </goals>
              <configuration>
                <arguments>run build</arguments>
                <environmentVariables>
                  <NODE_ENV>${NODE_ENV}</NODE_ENV>
                </environmentVariables>
              </configuration>
            </execution>
```
where NODE_ENV is a properties defined in pom properites section and overwritten by profiles of maven.

PR #454, opened 2016-08-23
Added documentation for environmentVariables

adumat:
After have watched the source code for find a method to pass some environment variables to node.js, i've noticed that this feature had been already implemented!
this pull request is only for adding the documentation about it and an example.

Reference to this ticket: #433


eirslett:
Thanks!

PR #455, opened 2016-08-24
reference scripts registered on node_modules/.bin

josmardias:
It helps in a lot of cases. For example, Grunt 1.0 contains grunt-cli, so we don't need to install grunt-cli alongside. In this case, refering node_modules/.bin/grunt will work no matter you are using grunt < 1.0 + grunt-cli, or just using grunt >= 1.0

### DIFF
--- a/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/BowerRunner.java
+++ b/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/BowerRunner.java
@@ -7,7 +7,7 @@ public interface BowerRunner extends NodeTaskRunner {}
 
 final class DefaultBowerRunner extends NodeTaskExecutor implements BowerRunner {
 
-    private static final String TASK_LOCATION = "node_modules/bower/bin/bower";
+    private static final String TASK_LOCATION = "node_modules/.bin/bower";
 
     DefaultBowerRunner(NodeExecutorConfig config, ProxyConfig proxyConfig) {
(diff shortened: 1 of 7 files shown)


josmardias:
WiP btw, the build is failing. Trying to figure out what is happening.
Just shared here to get any sugestions.

josmardias:
I'll squash those commits

josmardias:
The build works on my machine.
I tried `git clean -dfx` to remove all runtime files, and removed `~/.m2/repository/com/github/eirslett/` as well, but still works.

I have no idea what is happening to the CI

eirslett:
I'm not quite sure why it doesn't work. I vaguely remember trying to use node_modules/.bin a long time ago when the plugin was made, and for some reason, it didn't work, so it runs stuff inside node_modules/xxx/bin directly.

Does the build work everywhere? Travis runs the build in OS X, AppVeyor runs it on Windows, and CloudBees runs it on Linux.

josmardias:
node_modules/.bin files are symbolic links, maybe some issue with it?

josmardias:
I vaguely remember you saying something in another issue about this grunt/bower/jspm/whatever extensions being a mistake, and we should just use `npm scripts`. I agree with that.

Let's leave this PR open here for a while, just to guide people getting errors with the newer grunt 1.0 version.

Workaround:
For grunt@1.0, the workaround is either stop using the grunt support on this plugin and use npm directly running some `npm script`, or just leave grunt-cli installed as a npm direct dependency.

mriehema:
I agree. We switched to npm tasks too.

Issue #132, opened 2015-01-04
make the plugin threadSafe


royts:
hi,
when using maven to build multiple projects, you can ask to run the build in parallel by `-TX` when `X` is the number of threads.

when a plugin is not thread-safe the build will take more time to run.
from maven build output:
```
[WARNING] * Your build is requesting parallel execution, but project      *
[WARNING] * contains the following plugin(s) that are not marked as       *
[WARNING] * @threadSafe to support parallel building.                     *
[WARNING] * While this /may/ work fine, please look for plugin updates    *
[WARNING] * and/or request plugins be made thread-safe.                   *
[WARNING] * If reporting an issue, report it against the plugin in        *
[WARNING] * question, not against maven-core                              *
[WARNING] *****************************************************************
[WARNING] The following plugins are not marked @threadSafe in MyProject:
[WARNING] com.github.eirslett:frontend-maven-plugin:0.0.20
[WARNING] *****************************************************************
```

is `frontend-maven-plugin` thread-safe?
seems like [marking a plugin as thread-safe is easy](https://maven.apache.org/plugin-tools/maven-plugin-annotations/apidocs/org/apache/maven/plugins/annotations/Mojo.html#threadSafe()).
can we do that?

eirslett:
I'm not sure whether it's thread-safe or not... I guess that would be up to the build of the individual user. Some grunt builds could run in parallell, others would interfere with eachother. What's the common practice for mojos when you're not sure?

eirslett:
I'll close this due to inactivity.

Stephan202:
@eirslett, for a long time Maven stopped displaying this warning, but with [MNG-6171](https://issues.apache.org/jira/browse/MNG-6171) fixed and Maven 3.5.0 released last Friday, users building with `-T` will start seeing this warnings again.

In today's word of multi-core machines it's common to build a multi-module Maven build with `-T`. Have you since figured our whether the plugin is officially safe for concurrent execution? (Unofficially I can say it seems to always just have worked.)

Stephan202:
Doh, should have read #466 first. Well, if concurrent execution is now officially supported, then all that remains is to annotate the plugin accordingly :)

mriehema:
As you can see, I made` install-node-and-npm` threadsafe. Running npm, grunt, gulp or any other execution, they probably not threadsafe yet.

Stephan202:
[...]

PR #466, opened 2016-09-20
ThreadSafe Node and NPM installation

mriehema:
allow use of -T (threaded builds) together with node/npm installation

This is a copy of pull request #327 without proxy changes. We run into some race conditions in our CI with multiple modules using the frontend-maven-plugin with the same node installDirectory.


mriehema:
fixes issue #132 too.

eirslett:
Thanks!

PR #470, opened 2016-09-21
Check write permissions on node destination folder

### LINKED ISSUE #469: Cryptic error message when destination folder is not accessible
If the plugin fails to extract the node archive because it can't access the destination directory, the script will produce a somewhat cryptic error:

    [ERROR] Failed to execute goal com.github.eirslett:frontend-maven-plugin:1.0:install-node-and-npm
    (install node and npm) on project myproject: Could not extract the Node archive:
    Could not extract archive: '/home/user/.m2/repository/com/github/eirslett/node/4.4.5/node-4.4.5-linux-x64.tar.gz':
    No such file or directory

It's thrown by [`DefaultArchiveExtractor.extract():destPath.createNewFile();`](https://github.com/eirslett/frontend-maven-plugin/blob/master/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ArchiveExtractor.java#L43) which will produce the cryptic error message that can mislead the debugging process (at least I spent a decent amount of time trying to find out what was wrong with my node archive file)

    throw new ArchiveExtractionException("Could not extract archive: '"
                    + archive
                    + "'", e)

I can quickly add a PR to add a validation for this in the extract method, let me know if you want me to send it in.

jsvgoncalves:
* Raises a more detailed Exception if the plugin fails to extract the
  node archive because it can't access the destination directory.
  (Fixes #469)

Note: AccessDeniedException() was added since 1.7

### DIFF
--- a/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ArchiveExtractor.java
+++ b/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ArchiveExtractor.java
@@ -9,6 +9,7 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.nio.file.AccessDeniedException;
 
 class ArchiveExtractionException extends Exception {
     ArchiveExtractionException(String message, Throwable cause) {
@@ -39,6 +40,10 @@ final class DefaultArchiveExtractor implements ArchiveExtractor {
                 } else {
                     if (!destPath.getParentFile().exists()) {
                         destPath.getParentFile().mkdirs();
+                    }
+                    if(!destPath.getParentFile().canWrite()) {
+                        throw new AccessDeniedException(
+                                String.format("Could not get write permissions for '%s'", destinationDirectory));
                     }
                     destPath.createNewFile();
                     boolean isExecutable = (tarEntry.getMode() & 0100) > 0;
--- a/pom.xml
+++ b/pom.xml
@@ -60,8 +60,8 @@
                     <artifactId>maven-compiler-plugin</artifactId>
                     <version>3.1</version>
                     <configuration>
-                        <source>1.6</source>
-                        <target>1.6</target>
+                        <source>1.7</source>
+                        <target>1.7</target>
                     </configuration>
                 </plugin>
             </plugins>


jsvgoncalves:
@eirslett Please test

jsvgoncalves:
Looks like the build is passing now, as I changed the check to be on `!destPath.getParentFile()` rather than on `destPath`

mriehema:
@eirslett, could you check the appveyor settings for the windows build?

`Error: M2_HOME is set to an invalid directory. 
M2_HOME = "C:\Program Files (x86)\Apache\Maven" `

I got the same error in my pull request #466

eirslett:
> Note: AccessDeniedException() was added since 1.7

The project is still Java 6... But since even Java 7 has reached end-of-life, I think it's safe to go to 1.7. But the pom.xml files need to be updated! They state that 1.6 is the goal.

jsvgoncalves:
@eirslett the alternative is to use `FileNotFoundException` but I agree that it should be safe to go 1.7. So, should I add that change on the `pom.xml` to the PR? Even though I guess it would be good to have a separate PR for that.

eirslett:
Bumping it in the same PR is fine. (Maybe a separate commit, if you love rewriting git history then maybe a commit straight before this one, but it doesn't really matter)

jsvgoncalves:
@eirslett added the version bump commit

eirslett:
Ok, thanks for the PR!

mrothe:
We use this plugin for a while now and our builds are still (partial) on Java 6 (to be runtime compatible with older application servers). So what is wrong with a simple IOException(String.format("Could not get write permissions for '%s'", destinationDirectory));  ?

eirslett:
@mrothe old releases of the plugin will still work for you. I'll refrain from expressing my opinions on Java versions. :wink:

mrothe:
... or we use newer releases of the plugin by using maven toolchain on those build :-)
So I'm personally fine with Java 7. But is a AccessDeniedException (with string message) more meaningful than a IOException (with string message) to force Java 7 for a maven plugin? There are plenty of maven plugins out there which are compatible with Java 6 ... but again: This was just a comment and if this plugin is Java 7 starting from a 1.1 release I'm fine with that.

jsvgoncalves:
@mrothe I have no issues against reverting/rebasing the PR and change that into an IOException, but I think it's pointless. As @eirslett said, you can still use [version 1.0](https://github.com/eirslett/frontend-maven-plugin/releases/tag/1.0) and only the future versions (be it 1.1 or 2.0 or whatever) will require Java 7

PR #475, opened 2016-10-03
added support for newer versions of node that include npm

stevenbower:


eirslett:
Can you add an integration test that does this? (Similar to the other integrations tests that are already there)

stevenbower:
added integration in the same pattern as the existing that installs node 6.7.0 and used the "provided" NPM..

eirslett:
Looks like it failed on Windows...

frontend-maven-plugin/src/it/node-provided-npm/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+    <version>0</version>
+    <packaging>pom</packaging>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>com.github.eirslett</groupId>
+                <artifactId>frontend-maven-plugin</artifactId>
+                <!-- NB! Set <version> to the latest released version of frontend-maven-plugin, like in README.md -->
+                <version>1.1-SNAPSHOT</version>
```
mriehema: Please change the version back to @project.version@.

mriehema:
`NodeAndNPMInstaller `has `installNodeDefault()` and `installNodeForWindows()`. This pull request changes only the default case. :(

frontend-maven-plugin/src/it/node-provided-npm/pom.xml
eirslett: That's important, so that when we release 1.2, 1.3 etc. the integration tests will automatically use the current (snapshot) version of the plugin.

stevenbower:
Realized that the Windows path diverges from the Linux path as windows only ever try to download the single binary version.. I am almost done testing an update that will pull the zipped version on windows when npm is expected to be installed with node.. got it working but trying to minimize the amount of duplicated code.. 

Oh and yeh that 1.1-snapshot was in error.. will update the pom

mriehema:
Thx. That sounds good. But I have little request:
I just added a CHANGELOG.md. Could you merge/rebase your pull request and add a simple line to the changelog for this awesome enhancement? That would be nice.

stevenbower:
- Updated changelog
- Cleaned up POMs in test with proper version string
- windows support for packaged npm

README.md
```diff
@@ -21,170 +23,104 @@ laptops, but backend developers can run a clean build without even installing No
...
+        <goals>
+            <goal>install-node-and-npm</goal>
+        </goals>
+        <!-- optional: default phase is "generate-resources" -->
+        <phase>generate-resources</phase>
+    </execution>
     <configuration>
-        <workingDirectory>src/main/frontend</workingDirectory>
+        <nodeVersion>v4.6.0</nodeVersion>
+        <npmVersion>2.15.9</npmVersion>
```
mriehema: maybe you can add something like `<!-- optinal if nodeVersion >= 4 -->`

stevenbower: Will do

eirslett:
Looks much better now! Could you please rebase/squash the git history for the PR into one commit?

stevenbower:
squashed..

eirslett:
Thanks a lot!

PR #477, opened 2016-10-04
Documentation Optimzations

mriehema:
- added license
- added changelog
- optimized readme

I would suggest, that every new pull request should add one line to the changelog to keep maintenance low.


mriehema:
@eirslett, what do you think about changelog? If you want, I can take care about the maintenance.

eirslett:
That's great! Thanks a lot!

PR #480, opened 2016-10-05
Update example project

mriehema:
updated package.json dependencies to latest versions. changed node and npm to latest lts version. 
This should help testing current version instead of old legacy ones and it removes warnings from build log.


eirslett:
Thanks!

Issue #483, opened 2016-10-06
NoClassDefFoundError: org/codehaus/jackson/Versioned when running maven install


Geeman201:
Recently found this issue when running in an offline environment. Could just be the setup for the environment. Below is a section from the exception.

```
java.lang.NoClassDefFoundError: org/codehaus/jackson/Versioned
    java.lang.ClassLoader.defineClass1(Native Method)
    java.lang.ClassLoader.defineClass(ClassLoader.java:800)
```

I was using: 
 - Maven 3.3.9 
 - Java 1.8

My steps:
1. `mvn clean install` - works fine
2. `mvn install` - Fails with the above exception

Thought it was worth raising an issue in case anyone else has a similar issue. Adding the following fixed this for me:
```xml
<plugin>
    <groupId>com.github.eirslett</groupId>
    <artifactId>frontend-maven-plugin</artifactId>
    <version>1.0</version>
    <dependencies>
        <dependency>
            <groupId>org.codehaus.jackson</groupId>
            <artifactId>jackson-core-asl</artifactId>
            <version>1.9.13</version>
        </dependency>
    </dependencies>
</plugin>
```

Geeman201:
Raised a PR with an explicit dependency on `jackson-core-asl`

Looking in `jackson-mapper-asl` it is defined as a dependency in there so I'm unsure why this was failing for me.

PR #484, opened 2016-10-06
Fix issue with NoClassDefFound for org/codehaus/jackson/Versioned

Geeman201:
Reference to #483


eirslett:
Thanks for the fix!
The general code style for the pom.xml files in this project is to inline version numbers in the `<dependency>` declaration though, I think that's easier to work with.

Geeman201:
See update! I only extracted it as the Jackson version was the same!

eirslett:
Thanks!

Issue #490, opened 2016-10-12
Support for yarn


gunnarvelle:
Yarn package manager is a new tool wich can replace the npm client. A goal to run yarn would be nice.

https://yarnpkg.com

eirslett:
You could hack this together with the existing plugin. Do npm install yarn (not -g) with the npm mojo, and then have a package.json with (untested pseudocode)

```javascript
{
  scripts: {
    "yarn": "node node_modules/yarn/bin-something/yarn.js"
  }
}
```

And then, with the npm mojo again, do `npm run-script yarn`.

You could commit node_modules/yarn/* into your git repository if you want to skip the npm install yarn step.

It's a bit of work to implement first-class support for yarn. I suppose it's possible, if somebody is willing to take on the work.

marcins:
Ideally it would be first-class support as an alternative to npm rather than just a goal to run yarn - so you could specify which yarn version you want, and it'd be installed like npm gets installed as part of the `install-node-and-npm` goal (`install-node-and-yarn`? 😄 )

In our project we're currently blocked by yarn's lack of private registry support, but once we can use yarn I'll check back in to see where this issue is at. It's something I'd be potentially willing to help with once we need it!

mriehema:
Yarn looks promising. But it's in a very early stage. We should keep this ticket open and keep watching how yarn is developing. Nonetheless feel free to create a pull request for a `install-node-and-yarn` execution. But AFAIK yarn has just installer downloads (at least for Windows) and no simple executables, which are needed for this plugin.

mgohin:
With Yarn, my Jenkins Build (maven) from 1m35 to 25s, pretty nice. I had to remove this plugin to use a custom sh file with commands, I'd to put back this plugin with Yarn support ;)

mgohin:
Any news ? @maybeec made a PR

eirslett:
Just merged #499 which closes this issue.

mgohin:
A release soon ?

nanndoj:
+1 waiting for s release soon

eirslett:
v1.2 will be out now, once Maven central syncs with OSS Sonatype.

tatsujb:
where is this at?

the only people who are going to be using frontend-maven-plugin are employees working on a company project : thus behind proxy.

You can't get yarn via github.

this should be at least in the readme : 
in cmd or Powershell : (so that it at least works locally to boot and get you off the ground)
[...]

PR #492, opened 2016-10-14
first implementation yarn support

maybeec:
This is a PR for #490.
It just provides the first support for installing and running yarn. It has not yet been testing extensively, but embedded in a local build. It has been tested for windows only until now.

### DIFF
--- a/frontend-maven-plugin/src/it/custom-install-directory/verify.groovy
+++ b/frontend-maven-plugin/src/it/custom-install-directory/verify.groovy
@@ -1,2 +1,2 @@
-assert new File(basedir, 'target/node').exists() : "Node was not installed in the custom install directory";
+assert new File(basedir, 'target/.buildenv/node').exists() : "Node was not installed in the custom install directory";
 assert new File(basedir, 'node_modules').exists() : "Node modules were not installed in the base directory";
--- a/frontend-maven-plugin/src/it/yarn-integration/verify.groovy
+++ b/frontend-maven-plugin/src/it/yarn-integration/verify.groovy
@@ -1,5 +1,6 @@
 assert new File(basedir, 'target/.buildenv/node').exists() : "Node was not installed in the custom install directory";
 assert new File(basedir, 'node_modules').exists() : "Node modules were not installed in the base directory";
+assert new File(basedir, 'node_modules/less/package.json').exists() : "Less dependency has not been installed successfully";
 
 import org.codehaus.plexus.util.FileUtils;
 
--- a/frontend-maven-plugin/src/it/yarn-integration/verify.groovy
+++ b/frontend-maven-plugin/src/it/yarn-integration/verify.groovy
@@ -1,0 +1,8 @@
+assert new File(basedir, 'target/.buildenv/node').exists() : "Node was not installed in the custom install directory";
+assert new File(basedir, 'node_modules').exists() : "Node modules were not installed in the base directory";
+
+import org.codehaus.plexus.util.FileUtils;
+
+String buildLog = FileUtils.fileRead(new File(basedir, 'build.log'));
+
+assert buildLog.contains('BUILD SUCCESS') : 'build was not successful'
--- a/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/InstallNodeAndNpmMojo.java
+++ b/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/InstallNodeAndNpmMojo.java
@@ -1,45 +1,51 @@
 package com.github.eirslett.maven.plugins.frontend.mojo;
 
-import com.github.eirslett.maven.plugins.frontend.lib.FrontendPluginFactory;
-import com.github.eirslett.maven.plugins.frontend.lib.InstallationException;
-import com.github.eirslett.maven.plugins.frontend.lib.NodeAndNPMInstaller;
-import com.github.eirslett.maven.plugins.frontend.lib.ProxyConfig;
 import org.apache.maven.execution.MavenSession;
 import org.apache.maven.plugins.annotations.Component;
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
-import org.apache.maven.settings.crypto.SettingsDecrypter;
 import org.apache.maven.settings.Server;
+import org.apache.maven.settings.crypto.SettingsDecrypter;
 
-@Mojo(name="install-node-and-npm", defaultPhase = LifecyclePhase.GENERATE_RESOURCES, threadSafe = true)
+import com.github.eirslett.maven.plugins.frontend.lib.FrontendPluginFactory;
+import com.github.eirslett.maven.plugins.frontend.lib.InstallationException;
+import com.github.eirslett.maven.plugins.frontend.lib.NPMInstaller;
+import com.github.eirslett.maven.plugins.frontend.lib.NodeInstaller;
+import com.github.eirslett.maven.plugins.frontend.lib.ProxyConfig;
+
+@Mojo(name = "install-node-and-npm", defaultPhase = LifecyclePhase.GENERATE_RESOURCES, threadSafe = true)
 public final class InstallNodeAndNpmMojo extends AbstractFrontendMojo {
 
     /**
      * Where to download Node.js binary from. Defaults to http://nodejs.org/dist/
      */
-    @Parameter(property = "nodeDownloadRoot", required = false, defaultValue = NodeAndNPMInstaller.DEFAULT_NODEJS_DOWNLOAD_ROOT)
+    @Parameter(property = "nodeDownloadRoot", required = false,
+        defaultValue = NodeInstaller.DEFAULT_NODEJS_DOWNLOAD_ROOT)
     private String nodeDownloadRoot;
 
     /**
      * Where to download NPM binary from. Defaults to http://registry.npmjs.org/npm/-/
      */
-    @Parameter(property = "npmDownloadRoot", required = false, defaultValue = NodeAndNPMInstaller.DEFAULT_NPM_DOWNLOAD_ROOT)
+    @Parameter(property = "npmDownloadRoot", required = false,
+        defaultValue = NPMInstaller.DEFAULT_NPM_DOWNLOAD_ROOT)
     private String npmDownloadRoot;
 
     /**
      * Where to download Node.js and NPM binaries from.
      *
-     * @deprecated use {@link #nodeDownloadRoot} and {@link #npmDownloadRoot} instead, this configuration will be used only when no {@link #nodeDownloadRoot} or {@link #npmDownloadRoot} is specified.
+     * @deprecated use {@link #nodeDownloadRoot} and {@link #npmDownloadRoot} instead, this configuration will
+     *             be used only when no {@link #nodeDownloadRoot} or {@link #npmDownloadRoot} is specified.
      */
     @Parameter(property = "downloadRoot", required = false, defaultValue = "")
     @Deprecated
     private String downloadRoot;
 
     /**
-     * The version of Node.js to install. IMPORTANT! Most Node.js version names start with 'v', for example 'v0.10.18'
+     * The version of Node.js to install. IMPORTANT! Most Node.js version names start with 'v', for example
+     * 'v0.10.18'
      */
-    @Parameter(property="nodeVersion", required = true)
+    @Parameter(property = "nodeVersion", required = true)
     private String nodeVersion;
 
     /**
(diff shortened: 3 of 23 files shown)


eirslett:
Looks promising! Could you please add an integration test that installs node and yarn, then installs a couple of libs and runs some kind of test? Like the example project?

maybeec:
Yes, would be nice. I may have a look at this at the weekend if possible.

mriehema:
changelog and documentation is missing too.

maybeec:
where should I document anything? The existing code is missing documentation as well.

mriehema:
a new section in README.md to describe how to configure pom for yarn (should be similar to install-node-and-npm) and a simple one-liner in CHANGELOG.md should be enough for beginning.

eirslett:
The existing readme just points to the example project/integration tests - that way, the code is self-documenting and the documentation is always up-to-date. (At least in theory)

frontend-maven-plugin/src/it/yarn-integration/verify.groovy
```diff
@@ -0,0 +1,8 @@
+assert new File(basedir, 'target/.buildenv/node').exists() : "Node was not installed in the custom install directory";
+assert new File(basedir, 'node_modules').exists() : "Node modules were not installed in the base directory";
+
+import org.codehaus.plexus.util.FileUtils;
+
+String buildLog = FileUtils.fileRead(new File(basedir, 'build.log'));
+
+assert buildLog.contains('BUILD SUCCESS') : 'build was not successful'
```
eirslett: Maybe also assert that node_modules/less/package.json exists? Then we're sure the plugin has actually installed dependencies, not just short-circuited!

maybeec:
finally, got it to work also on linux. However, the download URL is just based on GitHub releases of Yarn. I did not found any webroot with all deployments. Hope, that the yarn guys will keep their naming conventions of the binaries consistent :)

I also adapted the test accordingly.

frontend-maven-plugin/src/it/custom-install-directory/verify.groovy
```diff
@@ -1,2 +1,2 @@
-assert new File(basedir, 'target/node').exists() : "Node was not installed in the custom install directory";
+assert new File(basedir, 'target/.buildenv/node').exists() : "Node was not installed in the custom install directory";
```
eirslett: I'm not sure it's such a good idea to add the .buildenv/ directory, since it will change the directory layout, which could hurt backwards compatibility. Maybe we can just have node/node(.exe), node/npm/xxx and node/yarn/xxx?

frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/InstallNodeAndNpmMojo.java
```diff
@@ -73,40 +79,38 @@ protected boolean skipExecution() {
...
         if (null != server) {
-            factory.getNodeAndNPMInstaller(proxyConfig)
-                .setNodeVersion(nodeVersion)
-                .setNodeDownloadRoot(nodeDownloadRoot)
-                .setNpmVersion(npmVersion)
-                .setNpmDownloadRoot(npmDownloadRoot)
-                .setUserName(server.getUsername())
-                .setPassword(server.getPassword())
-                .install();
+            factory.getNodeInstaller(proxyConfig).setNodeVersion(this.nodeVersion)
```
eirslett: Why has this stuff been changed? It's not necessary.

frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/InstallNodeAndYarnMojo.java
```diff
@@ -0,0 +1,89 @@
...
+        return this.skip;
+    }
+
+    @Override
+    public void execute(FrontendPluginFactory factory) throws InstallationException {
+        ProxyConfig proxyConfig = MojoUtils.getProxyConfig(this.session, this.decrypter);
+        Server server = MojoUtils.decryptServer(this.serverId, this.session, this.decrypter);
+        if (null != server) {
+            factory.getNodeInstaller(proxyConfig).setNodeDownloadRoot(this.nodeDownloadRoot)
+                .setNodeVersion(this.nodeVersion).setPassword(server.getPassword())
```
eirslett: It's much easier to read if there is one method call per line.

frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ArchiveExtractor.java
```diff
@@ -29,86 +34,101 @@
...
     public void extract(String archive, String destinationDirectory) throws ArchiveExtractionException {
         final File archiveFile = new File(archive);
-        FileInputStream fis = null;
-        try {
-            fis = new FileInputStream(archiveFile);
 
-            if("zip".equals(FileUtils.getExtension(archiveFile.getAbsolutePath()))){
+        try (FileInputStream fis = new FileInputStream(archiveFile)) {
+            if ("msi".equals(FileUtils.getExtension(archiveFile.getAbsolutePath()))) {
+                String command = "msiexec /a " + archiveFile.getAbsolutePath() + " /qn TARGETDIR=\""
```
eirslett: Will this cause the yarn installation to be registered globally in the system, even if the files are placed in the local folder? The operating system shouldn't know about yarn being installed (because then it can be accidentally uninstalled).

frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/FrontendPluginFactory.java
```diff
@@ -3,34 +3,45 @@
 import java.io.File;
 
 public final class FrontendPluginFactory {
-    
+
```
eirslett: It's not necessary to fix all code style things here, adding new methods is enough. (I suppose code style fixes could be a separate PR)

frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NodeExecutorConfig.java
```diff
@@ -3,43 +3,48 @@
...
+    @Override
+    public File getNpmPath() {
+        return new File(this.installConfig.getInstallDirectory() + Utils.normalize(NPM));
+    }
 
-  @Override
-  public File getWorkingDirectory() {
-    return installConfig.getWorkingDirectory();
-  }
+    @Override
```
eirslett: Fixing the code style can be done in a separate PR (to keep the line change count down on this PR)

frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/YarnTaskExecutor.java
```diff
@@ -0,0 +1,123 @@
...
+            final boolean maskMavenProxyPassword = s.contains("proxy=");
+            if (maskMavenProxyPassword) {
+                final String bestEffortMaskedPassword = maskPassword(s);
+                clonedArguments.set(i, bestEffortMaskedPassword);
+            }
+        }
+        return "'" + taskName + " " + implode(" ", clonedArguments) + "'";
+    }
+
+    private static String maskPassword(String proxyString) {
```
eirslett: I guess this function could be moved into a separate helper class somewhere? I suspect it's used somewhere else too.

frontend-maven-plugin/src/it/custom-install-directory/verify.groovy
maybeec: basically, it is your decision. However, the name of the temporary directory `node` is not a good choice especially, if you want to ignore it on commit. Therefore, I introduced a more speaking and generic folder name to also include further tmp local environments in future.
I even do not see any "backward-compatibility issues" here as this is just a folder of temporary files, which will even be recreated at build time, but correct me if I am wrong.

So I would highly recommend to change the folder name.

frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/InstallNodeAndNpmMojo.java
maybeec: It has just been changed due to IDE auto-formatting reasons. Basically, I was lazy to format code by myself. However, I would guess that you have your own favorite code style, which is not shared anyhow? I am applying my code styles while merging before doing a review. Thus, I can see what actually has been changed and merge the code with the correct code style.

frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/InstallNodeAndYarnMojo.java
maybeec: personal preference, but here just as of my used IDE settings.

frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ArchiveExtractor.java
maybeec: As far as I got to know, the msiexec command just extracts the archive to a given target folder. There is no software registry updated.

frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/YarnTaskExecutor.java
eirslett: Yes, it's right that the visibility stuff is written a bit weird, it should probably be changed so all classes are visible, at least in their own package. I suppose we can skip refactoring this one then, and to it later :-)

frontend-maven-plugin/src/it/custom-install-directory/verify.groovy
mriehema: Please add a separate PR for this change to discuss. This has nothing to do with yarn and would break a lot of existing projects.

frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/InstallNodeAndNpmMojo.java
```diff
@@ -1,45 +1,51 @@
...
 import org.apache.maven.plugins.annotations.Parameter;
-import org.apache.maven.settings.crypto.SettingsDecrypter;
 import org.apache.maven.settings.Server;
+import org.apache.maven.settings.crypto.SettingsDecrypter;
+
+import com.github.eirslett.maven.plugins.frontend.lib.FrontendPluginFactory;
+import com.github.eirslett.maven.plugins.frontend.lib.InstallationException;
+import com.github.eirslett.maven.plugins.frontend.lib.NPMInstaller;
+import com.github.eirslett.maven.plugins.frontend.lib.NodeInstaller;
+import com.github.eirslett.maven.plugins.frontend.lib.ProxyConfig;
```
mriehema: Please revert all changes, which are not needed for yarn. That would help us to review your PR.

frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ArchiveExtractor.java
mriehema: I will test that on my Windows machine.

mandragorn:
Is there anything I can do to help out with this?  I'm actively looking at switching our organization over to yarn and we currently use the frontend-maven-plugin.  Until this is merged I'll be using the workarounds described in #490

maybeec:
Actually, the only open issue here is the code formatting as far as I can see? Right?
As I am currently running out of time, I will not be able to rework the code, I can give anyone access to my fork if interested. However, I would recommend to the plugin authors to do it by their own if they really want to do so. 
By the way: From my experiences, GitHub UI seems to not be able to calculate the final diff of the PR and the master branch. It will moreover create a view by just aggregating all changes, which leads you to review changes, which have been eliminated by later commits on the PR. So I would recommend to just merge and review on the local machine.
Then review might be even easier.

maybeec:
@mriehema can you argument on your comment? What should break for already existing projects, when using a different tmp folder?
I would even argument against it and say, that the current folder "node" is a bad choice regarding name. The name does not include any of its original semantics and it might be even worse when trying to exclude it recursively from version control systems as "node" might be an often used folder name in code.

maiis:
Looks awesome. Any plan to merge this feature ? Can't wait 👍 💯

eirslett:
There are some concerns that have to be addressed before it is merged (read the thread), I suspect @mandragorn or @mriehema are willing to take a look at them?

polarctos:
I addressed all three concerns in https://github.com/polarctos/frontend-maven-plugin/commit/c2c053323cd7dc68ff714d4509cbe0d76bdb665d:
* reverted the formatting changes
* removed usage of the .buildenv folder
* added a section to the readme file

Is there anything else that is missing to get this squashed and merged?

eirslett:
Thanks a lot! That should be okey then! Can you squash everything into one commit and make a new PR to this repo? Then I'll give it another review and merge it finally :-)

eirslett:
This PR was updated in #499 and merged, so I'll close this issue.

PR #496, opened 2016-10-20
Change Travis Configuration

mriehema:
- add linux
- switch to jdk8
- change language

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ language: java
 matrix:
   include:
     - os: linux
-      jdk: oraclejdk8
+      jdk: oraclejdk7
     - os: osx
-      osx_image: xcode8
+      #osx_image: beta-xcode6.1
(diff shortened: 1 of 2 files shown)


eirslett:
Is it possible to use Java 7?

mriehema:
Sure, if you want, even both JDKs.

eirslett:
It's enough with only Java 7, then the build will be faster as well.

mriehema:
Travis changed the default JDK for OSX to Java 8. Its not possible to change it. See https://github.com/travis-ci/travis-ci/issues/6716 All I can do, is switch back to an old OSX version 10.9. I would prefer using JDK8 with a new OSX.

mriehema:
@eirslett Something missing or wrong for a merge?

eirslett:
Oh sorry, I thought this was already merged. Thanks! Will merge it now.

PR #500, opened 2016-10-31
Fix Yarn version comparison

cdarras:
The command `yarn --version` doesn't return a version number prefixed with a **v**.


eirslett:
Thanks!

Issue #506, opened 2016-11-07
The skip parameter on InstallNodeAndYarnMojo is ignored


alexcojocaru:
**Do you want to request a _feature_ or report a _bug_?**
BUG

**What is the current behavior?**
Setting the skip parameter (skip.installyarn) to true in the configuration on InstallNodeAndYarnMojo does not seem to have any effect - the mojo still executes.

**If the current behavior is a bug, please provide the steps to reproduce.**
Create an empty maven project:
`$ mvn archetype:generate -DgroupId=com.github.alexcojocaru -DartifactId=frontend-maven-plugin-test -DarchetypeArtifactId=maven-archetype-quickstart -DinteractiveMode=false`

Change the pom.xml to look like: https://gist.github.com/alexcojocaru/2f24d4f900a9a5fe91121f6a82377db7

**What is the expected behavior?**
Running `$ mvn generate-resources` on the project should skip the installation of node and yarn.

**Please mention your frontend-maven-plugin and operating system version.**
1.3-SNAPSHOT / OSX / Maven 3.2.3

This change seems to fix the issue; since I don't understand why it fixes the problem, I would rather not submit a merge request yet.

```diff --git a/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/InstallNodeAndYarnMojo.java b/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/InstallNodeAndY
index cab3a2d..c609e90 100644
--- a/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/InstallNodeAndYarnMojo.java
+++ b/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/InstallNodeAndYarnMojo.java
@@ -56,7 +56,7 @@ public final class InstallNodeAndYarnMojo extends AbstractFrontendMojo {
     /**
      * Skips execution of this mojo.
      */
-    @Parameter(property = "skip.installyarn", defaultValue = "false")
+    @Parameter(property = "skip.installyarn", alias = "skip.installyarn", defaultValue = "false")
     private Boolean skip;

     @Component(role = SettingsDecrypter.class)```

eirslett:
If you `mvn generate-resources -Dskip.installyarn=true`, on the command line then it works. I don't think we support disabling it in the pom.xml (yet).

[...]

PR #507, opened 2016-11-08
fix the skipping of the install-node-and-yarn execution in pom

alexcojocaru:
The issue fixed by this changeset is described in https://github.com/eirslett/frontend-maven-plugin/issues/506.
In summary, the execution of the install-node-and-yarn mojo cannot be skipped through the config in the pom.xml file.


eirslett:
Thanks!

PR #510, opened 2016-11-11
Possibility define already installed yarn for plugin

hadadil:
**Summary**

We have pre-installed `yarn` in our environments where the folder structure not matching with that which is defined in the plugin at the moment. Would be nice to have some possibility to define a custom path of yarn executable. This change is implement this need.

Also would be nice to influence this path by environment variable somehow.
So basically all environment would know is there already installed instance of yarn.
Or the plugin also would check the `PATH` to see is already on it.


eirslett:
I'm sorry, I will not merge this pull request, because it goes against the design of the plugin. Every build dependency/tool (node.js, npm, yarn) is supposed to be installed locally. (It doesn't matter if yarn is also installed globally, they will not interfere)
If you insist to use the system version of yarn, you should use something like exec-maven-plugin instead.

Issue #503, opened 2016-11-02
Yarn - windows tarball support & custom installation path


kash-raman:
1. **Windows without admin access** By default in windows, plugin is trying to install yarn msi. With a standard user, I'm not able to install. 
Need a toggle in YarnInstaller.java line no. 127 to switch using tar.gz.  

2. **Custom path support:** Due to above issue, I'm using below execution for installation and need it in global scope. The installation happens under working_folder/node/node_modules/yarn/... which is totally different folder from defaulted target/node/yarn.  And it does not honor workingDirectory or installedDirectory.

```
                  <execution>
                        <id>npm install yarn</id>
                        <phase>prepare-package</phase>
                        <goals>
                            <goal>npm</goal>
                        </goals>
                        <configuration>
                            <arguments>install grunt/yarn-complete.tgz --verbose -g</arguments>
                            <skip>${skipGruntInstall}</skip>
                        </configuration>
                    </execution>

```
```
public class YarnInstaller {
    public static final String **INSTALL_PATH = "/node/yarn"**;
```

```
final class InstallYarnExecutorConfig implements YarnExecutorConfig {
    private static final String YARN_WINDOWS =
        YarnInstaller.INSTALL_PATH.replaceAll("/", "\\\\") + "\\Yarn\\bin\\yarn.cmd";
```

 
Is anyone facing similar issue? Also, I'm working on the fix.
 
Thanks,
Kasthuri

eirslett:
We should not install yarn globally (that defeats the point with the plugin). But it's strange that the msi creates so much trouble... maybe it only installs programs globally after all?

@maybeec @mriehema Have you experienced this?

mriehema:
Sorry, I hadn't time to test it yet (had some days off). So you have been faster with merging and releasing it. ;) I will test it ASAP.

maybeec:
As far as I understood msi, it will register the installed software globally into the software registry of windows when running the installation properly. What I have done in the yarn deployment in my PR was just extracting the msi, which should skip the software registration. However, due to this issue right management seem to intercept the extraction as well which is a pitty. I am not sure how to circumvent this. I am not sure, maybe I can spend an hour googling at the weekend :)

mriehema:
So, here are my retest results:

Run in git bash and cmd:
Variant 1: No yarn, just frontend-maven-plugin with default yarn settings without admin access: :ok: 
[...]

PR #512, opened 2016-11-14
Windows  support

kash-raman:
**Summary**
Windows MSI file requires registry change. Switching it to use tarball.  https://github.com/eirslett/frontend-maven-plugin/issues/503


eirslett:
Thanks!

PR #517, opened 2016-11-17
Added bower proxy ignore parameter

egavard:
**Summary**

The aim of this modification is to allow user to NOT use maven proxy settings on bower commands.

**Tests and Documentation**

This modification is sort of a copy of the existing code for Node proxy ignoring method.


jkokora:
Hope this is pulled soon as we cannot take this plugin into use due to this implementation missing.

eirslett:
Sorry for the delay. Thanks!

PR #520, opened 2016-11-21
[WiP] It's time for a 2.x

josmardias:
Work in progress PR. Please, do not accept/reject it.

This is a first step to a 2.x release.
I'm opening this PR to collect opinions from the community.

(discussion moved to josmardias/frontend-maven-plugin/issues/1)


eirslett:
Thanks for the initiative!
One thing I have been thinking about for a possible 2.x release, is to remove the implementation entirely and use [nvm](https://github.com/creationix/nvm) and [nvmw](https://github.com/hakobera/nvmw) (or [nvm-windows](https://github.com/coreybutler/nvm-windows) or another implementation) instead. Then, any improvements in nvm would benefit this plugin, and the other way around.

I think we should remove support for having a custom npm, just always use the provided npm. Using another version isn't really a use case, it's more of a hassle to figure out which versions match.

> downloadUrls should accept array of <url>'s

This one is problematic since the URLs would be different per platform. I was thinking we could just go with (node/npm/yarn)downloadRoot and then something like repositoryLayout which defaults to "nodejs" or "npmjs" but can also be set to "nexus", that layout would be a function from application + version to URL path.

The biggest problem with starting a 2.x line is the amount of work, which is really a lot more than you would think. Documentation must be updated, there must be migration guides, beta releases etc., and somebody will always complain that there are features from 1.x missing and we must have guidelines pointing them to the 1.x release line. In order to justify that amount of work there has to be very good reasons, and I'm not sure simplifying the API itself is reason enough.

josmardias:
tl;dr: 
Sorry about this long answer. I'm not trying to convince you by throwing a huge amount of information. In fact, I understand your point about the "amount of work, which is really a lot more than _I_ would think", and I agree with that. My problem is that I'll have to work on _Node download supporting fallbacks and maven settings <server> credentials_ and on _support multiple NPM registries because NPM itself don't_ anyway, because I need it working. It could be a _way better_ for me to work on it supporting a minimalist API. That's why I want a 2.x release with those features :dancer:.

> I think we should remove support for having a custom npm

I was about to propose this one, but the Yarn integration almost convinced me (there is a "version" free Yarn, why not NPM?).

> This one is problematic since the URLs would be different per platform.

I proposed to change `downloadRoot` to something else, because I always think it's the `local file system download root`, where all downloaded versions goes. Seems like you already put some thoughts on this, and `downloadUrls` is not ok as well.
My intention here was not to change the property behavior, only it's name. We should keep expecting a mimic of Node.js dist/ folder structure, for simplicity, without supporting other structures.
Actually, Nexus does not advocate for Node.js /dist structure. You just configurate a blob proxy for `https://nodejs.org/dist/`, and it just works.

> I'm not sure simplifying the API itself is reason enough.

I think there is some kind of fatigue here about documenting this project. "We" spend a lot of time on documentation, and tomorrow there is _another npm project pull request_, that works, but doesn't have any documentation, and people start yelling at you to accept it. Without clear bounds it will only keep growing to the wrong direction.
I thought we could try to fix this fatigue moving to a minimalist API with a very clear position about PR's we are _not_ accepting.

I think we should start a 2.x branch (actually, moving 1.x to a branch and starting 2.x on master) because  documentation could be a lot simpler.

For example:
```
<!-- I think it's easier to document and and have a cleaner code if we only support this structure -->
<downloadRoot>
  <!-- try to download from here first -->
  <url>
    <id>http://myproxy.example.org/nodejs/</id>
    <!-- it's a <server> credential on maven settings -->
    <serverId>download-node</serverId>
  </url>
  <!-- Node official registry is used if no downloadRoot is provided, so it needs to be here if you want a fallback -->
  <!-- It helps us not having to create a "do not download from official" option in the future -->
  <url>https://nodejs.org/dist/</url>
</downloadRoot>

<!-- and do not support this one anymore -->
<!-- we just can't explain how downloadRoot works in a single configuration example -->
<downloadRoot>http://myproxy.example.org/nodejs/<downloadRoot>
```

Another example:
```
<!-- a single example to know how to install node, and it's package managers -->
<plugin>
    <execution>
        <!-- <id> is optional, for better logging -->
        <goals>
            <!-- a single goal to download node, and possibly a custom npm or yarn -->
            <goal>download</goal>
        </goals>
        <!-- <phase> defaults to "generate-resources" -->
    </execution>
    <configuration>
        <!-- supports only node v4 or greater -->
        <version>v4.6.0</version>

        <!-- optional, NPM is already provided within Node.js installer -->
        <npm>
          <version>2.15.9</version>
          <downloadRoot>
            <!-- try to download from here first -->
            <url>
              <id>https://myproxy.example.org/npm/</id>
              <!-- it's a <server> credential on maven settings -->
              <serverId>download-npm</serverId>
            </url>
            <!-- NPM official registry is used if no downloadRoot is provided, so it needs to be here if you want a fallback -->
            <!-- It helps us not having to create a "do not download from official" option in the future -->
            <url>http://registry.npmjs.org/npm/-/</url>
          </downloadRoot>
        </npm>

        <!-- optional, to download Yarn as alternative -->
        <yarn>
          <version>v0.16.1</version>
          <downloadRoot>
            <!-- try to download from here first -->
            <url>
              <id>http://myproxy.example.org/yarn/</id>
              <!-- it's a <server> credential on maven settings -->
              <serverId>download-yarn</serverId>
            </url>
            <!-- Yarn official registry is used if no downloadRoot is provided, so it needs to be here if you want a fallback -->
            <!-- It helps us not having to create a "do not download from official" option in the future -->
            <url>https://github.com/yarnpkg/yarn/releases/download/</url>
          </downloadRoot>
        </yarn>

    </configuration>
</plugin>
```

eirslett:
> I think there is some kind of fatigue here about documenting this project. "We" spend a lot of time on documentation, and tomorrow there is another npm project pull request, that works, but doesn't have any documentation, and people start yelling at you to accept it. 

This sums up the project pretty well. I consider this maven plugin feature complete. It does what it promises (and is intended for), it doesn't promise anything that it doesn't support.

It was a weekend project, and was completed around version 0.0.5, when it covered my former employer's needs. It was never intented to be as popular as it became. Everything that has happened since then (except for the late yarn support) is small tweaks and improvements. 5 % of its features will work for 95 % of developers, and the other 95 % of features will enable 98 % of developers to use it.

It's always a problem that people want open source projects to support every possible feature under the sun, and it's a real challenge to set these boundaries. (Just look at the number of closed issues/PRs for supporting system provided node/npm) I tried to limit it, but gave up eventually, because that is tiring as well.
If it were up to me, I wouldn't even support maven proxy integration. I would try to offload that kind of work to other projects that implement it better, like nvm.
I don't even use this maven plugin anymore, I work on pure Node.js projects. So in my free unpaid time, I sometimes look at issues, merge PRs and make new releases.

I'm open to giving access to the git repo for people who are willing to make an effort to develop the plugin further, but it would take at least 3 motivated developers or so, to champion it, and maintain a 2.x release line in the longer term (look at PRs and issues). There are already a number of developers who have helped out a lot with fixes/improvements and helping other people with issues, I greatly appreciate that!

It would of course require a general agreement on what the right direction is for the project, going forward. What should be supported, what should not, how should problems best be solved. Who should be denied support, based on too-specific requirements. These are difficult questions.

josmardias:
Discussion moved to josmardias/frontend-maven-plugin/issues/1
I'll work on another branch, and probably a separated CI.
I'll work on this changes anyway because I need them.

Feel free to reopen the discussion anytime,
Thanks for this great plugin!

Closing the PR.

PR #535, opened 2016-12-11
Document how to skip package managers and build tools

matthew-dailey:
**Summary**

In modules containing both Java and frontend code, it is nice to sometimes skip the frontend build/test process.  This functionality already exists, so I wanted to document it in the README.

* Listed out build tools and skip parameters
* Notice about skipping package managers requires also
skipping build tools
* Did _not_ list `-Dskip.ember` because the rest of the ember support is not discussed in the README


eirslett:
Thanks!

PR #550, opened 2017-01-10
Feature/protractor

ningod:
**Summary**

* Added new goal `protractor` for Protractor support

Protractor is an end-to-end test framework for AngularJS applications. Protractor runs tests against your application running in a real browser, interacting with it as a user would.

**Tests and Documentation**

* Changed README.md adding protractor's informations
* Added tests on `example project`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Changelog
 
+### 1.4
+
+* New goal `protractor` for Protractor support 
+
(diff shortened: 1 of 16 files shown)


ningod:
The only problem in appveyor build is due to maven installation on their system

`Failures: maven (exited -1) - Error while running C:\ProgramData\chocolatey\lib\maven\tools\chocolateyInstall.ps1.`

PR #563, opened 2017-02-02
Add the ability to run node with additional parameters i.e. --max-old…

HearnMichael:
…-space-size=4000

This feature provides the ability to pass additional arguments to node when running the different node tasks.

I needed the ability to increase the max-old-space-size node was using when running webpack.  This feature allows these kinds of extra arguments to be passed when executing node tasks.

**Tests and Documentation**

Updated README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ to see how it should be set up: https://github.com/eirslett/frontend-maven-plugi
     - [Proxy Settings](#proxy-settings)
     - [Environment variables](#environment-variables)
     - [Skipping Execution](#skipping-execution)
+    - [Node arguments](#node-arguments)
     
 **Recommendation:** _Try to run all your tasks via npm scripts instead of running bower, grunt, gulp etc. directly._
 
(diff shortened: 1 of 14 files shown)


MastaP:
any chances to get this PR merged any time soon?

eirslett:
This is an edge case. Is it possible to use the `node` task directly and provide this flag in `<arguments>`?

MastaP:
We're using `yarn` goal, not `node` directly

eirslett:
You could use the `node` goal to call `yarn` right?

MastaP:
yes but what's the point of a dedicated `yarn` (or any other non `node`) goal is if everything could be run using `node`?

eirslett:
It's mostly for backwards compatibility; if I could have written this plugin again from scratch, it would only contain the `node` goal.

MastaP:
okay, in that case, if you really don't want community to help you out with improving the plugin (I'm looking at 176 open issues and 46 pull requests), at least close stuff do don't want to merge. keeping this PR open for almost 2 years without any feedback is not polite. thanks bye

PR #553, opened 2017-01-11
Added maven.frontend.failOnError and maven.test.failure.ignore support

ningod:
**Added flags to best manage integration-test**

This pull request resolve the need to have some configuration to ignore failures during build, expecially to manage maven integration-test phases (pre and post)  as explained in: https://github.com/eirslett/frontend-maven-plugin/issues/552

PR #564, opened 2017-02-10
closes #534: Do not fail the Maven build for a failed test

### LINKED ISSUE #534: Feature request: Do not fail the Maven build for a failed test
When running `npm test` (for example) and if it fails, the whole Maven build will fails.
Would it be possible to make the plugin act differently for goals invoked during `test` or `integration-test` phases so that failures in those are only reported as test failures?

If you can provide some guidance on whether this can be achieved and how, I'd gladly help with implementing it.

kaqqao:
Added general support for `testFailureIgnore` (`maven.test.failure.ignore`) flag as in Surefire plugin (this closes #534). Currently only Karma mojo respects this flag, making it impossible to apply it to Protractor or any other tests.
In case this flag is set to `true`, a failure during `test` or `integration-test` phase will not fail the entire build, but just log the error.

**Property name** (to add to the `<configuration>` block): `testFailureIgnore`
**User property** (to add as command line argument, or the `<properties>` block): `maven.test.failure.ignore`

The change is minimal, and the flag itself, the behavior, and the implementation are all consistent with [Surefire plugin](https://github.com/apache/maven-surefire/blob/303cc8acdaf677bee9751da94de16e07083febb7/maven-surefire-common/src/main/java/org/apache/maven/plugin/surefire/SurefireHelper.java):

**EDIT**
Seems like a similar pull request already exists: https://github.com/eirslett/frontend-maven-plugin/pull/553

### DIFF
--- a/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/AbstractFrontendMojo.java
+++ b/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/AbstractFrontendMojo.java
@@ -27,6 +27,13 @@ public abstract class AbstractFrontendMojo extends AbstractMojo {
   protected Boolean skipTests;
 
   /**
+   * Set this to "true" to ignore a failure during testing. Its use is NOT RECOMMENDED, but quite convenient on
+   * occasion.
+   */
+  @Parameter(property = "maven.test.failure.ignore", defaultValue = "false")
+  protected Boolean testFailureIgnore;
+
+  /**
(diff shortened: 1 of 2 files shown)


qacwnfq:
Please merge this, it's just what I need! :+1:

mriehema:
Looks like this PR, PR #572 and PR #553 fix the same My favorite is #572 because it includes documentation and release notes. 
@kaqqao Could you check, if PR #572 works for you? If yes, could you close your PR?

kaqqao:
Yes, I'd be quite happy to get _any_ of the PRs merged. Not particularly attached to mine. #572 seems good.

Issue #552, opened 2017-01-11
Request Feature - Add failsafe mode to generally manage integration-test


ningod:
In my project i use frontend maven plugin to execute integration-test with protractor testing framework, in my configuration i launch protractor using yarn goal and a script in package.json , but if tests fails all build fails and post-integration-test phase is never executed.
Due to this I need something to ignore plugin failure, a failOnError property or something similar to karma behavior with the "maven.test.failure.ignore" property

eirslett:
We do have that property for the karma mojo, I think. I just cannot remember what it's called. Check the examples, or maybe in the code!

ningod:
Yes, we have in karma but I'm not executing karma tests... I need it in other goals.
I'm moving the property from karma mojo to parent class then I'll pull a request to you

kaqqao:
This duplicates my issue https://github.com/eirslett/frontend-maven-plugin/issues/534
but the related pull request seems to have more features than mine (plus examples, it seems), so I'd favor this one over mine.

PR #572, opened 2017-02-20
add failOnError and integration-test support

jcarsique:
This reuses and fixes PR #553 , solving issue #552 ; see also #534 closed as duplicate.

Current behavior: an execution failure stops the Maven build. This is an issue for instance when the plugin is used to execute tests. Moreover, after an `integration-test` phase, it is often required to execute the `post-integration-test` phase for environment cleanup and tear down.

Expected behavior: 
- add a `failOnError` property
- if unset, then defaults to true except if if the current execution is a testing phase
- else, if `failOnError==true`, keep the current behavior (`BUILD FAILURE`)
- else, if `failOnError==false`, don't fail the Maven build (`BUILD ERROR`)

That is compliant with the common behavior encountered on the Maven plugins.

Tested on frontend-maven-plugin 1.4-SNAPSHOT

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ target
 .project
 .metadata
 .recommenders
+bin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Changelog
 
+### 1.4
+
+* Add maven.frontend.failOnError and maven.test.failure.ignore flags to best manage integration-test
+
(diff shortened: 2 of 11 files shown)


mriehema:
LGTM, what do you think @eirslett?

eirslett:
Ok thanks, I'll merge this version!

eirslett:
This PR was reverted due to #613.

Issue #595, opened 2017-05-07
[bug] Skipping goals -Dskip.npm and -Dskip.webpack doesn't work


AndreyRubankov:
As a developer I want to be able to skip _npm install_ step through -Dskip.npm=true.

Steps to reproduce:
1. Run the command: mvn clean install -Dskip.npm=true -Dskip.webpack=true

Expected: The npm step and the webpack step will be skipped.
Actual: Both steps run.

Plugin Versions: 1.3, 1.4, 1.5-SNAPSHOT
OS: MacOs
Java1.8
Maven: 3.2.5, 3.5.0

AndreyRubankov:
I've investigated this issue and found when this issue can be fixed.

It's looks pretty simple: change default value from hardcoded 'false' to EL tag with property name.
```
@Parameter(property = "skip.npm", defaultValue = "${skip.npm}")
private boolean skip;
```
In my local tests it was fixed this issue.

@eirslett , If you agree with this solution I'll create patch little bit later.

eirslett:
Yes, that sounds like a good idea.

AndreyRubankov:
PR for fix has been created: https://github.com/eirslett/frontend-maven-plugin/pull/596 .

Could you please review it and merge?

AndreyRubankov:
Thank you Eirik,
So, for history: this issue will be fixed in version 1.5.

PR #596, opened 2017-05-07
595: Fixed issue with skip properties with jvm arguments

AndreyRubankov:
Fix for issue [https://github.com/eirslett/frontend-maven-plugin/issues/595](url).

**Summary**
All hardcoded "false" values replaced with EL code for each skip-property.
`@Parameter(property = "skip.npm", defaultValue = "${skip.npm}")`
This value will be read from jvm argumens like -Dskip.npm=true

Also changed type of skip - all big Boolean replaced with little boolean, because big Boolean can be null. In case when skip value set with null the `boolean skipExecution` method will throw NullPointerException (implicit unboxing) and build will be failed:
> Execution npm install of goal com.github.eirslett:frontend-maven-plugin:1.5-SNAPSHOT:npm failed.: NullPointerException


eirslett:
Thanks!

PR #616, opened 2017-06-27
Update the default location of npm from http://registry.npmjs.org/npm/-/ to https://registry.npmjs.org/npm/-/ .

dbaxa:
Instead of using http://registry.npmjs.org/npm/-/  use https://registry.npmjs.org/npm/-/ by default. 
As per, https://github.com/eirslett/frontend-maven-plugin/pull/617, 

> Summary
>
> It is good security practice to download via HTTPS over HTTP where available, this change updates the default URL to download NPM to be HTTPS not HTTP. Node is already downloaded over HTTPS.
>
>Tests and Documentation
>
>The README, and various comments, have been updated with the correct default URL (also there was one instance of a NodeJS download URL still mentioning HTTP)

Signed-off-by: David Black <dblack@atlassian.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ You can also specify separate download roots for npm and node as they are stored
     <configuration>
         <!-- optional: where to download node from. Defaults to https://nodejs.org/dist/ -->
         <nodeDownloadRoot>http://myproxy.example.org/nodejs/</nodeDownloadRoot>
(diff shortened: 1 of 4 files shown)


dbaxa:
cc @eirslett

eirslett:
thanks!

PR #617, opened 2017-06-27
use HTTPS instead of HTTP to download NPM

marcins:
**Summary**

It is good security practice to download via HTTPS over HTTP where available, this change updates the default URL to download NPM to be HTTPS not HTTP.  Node is already downloaded over HTTPS.

**Tests and Documentation**

The README, and various comments, have been updated with the correct default URL (also there was one instance of a NodeJS download URL still mentioning HTTP)


marcins:
Never mind, #616 already does this.

PR #621, opened 2017-07-05
Enable proxy overrides for install-node-and-npm

bradfordwagner:
In enterprise systems it is sometimes not possible to change settings.xml in order to set proxy information. In this case it makes it difficult to download node/npm. This change allows a user to specify a download configuration for goal: install-node-and-npm.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ Last public release: [![Maven Central](https://maven-badges.herokuapp.com/maven-
 
 ## Changelog
 
+### 1.5
+* Add proxy override configuration for `install-node-and-npm`
+
 ### 1.4
 
 * Add maven.frontend.failOnError and maven.test.failure.ignore flags to best manage integration-test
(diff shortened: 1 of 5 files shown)


bradfordwagner:
@eirslett Would someone be able to review this PR please?

eirslett:
It seems like an overly complicated solution to a somewhat edge-casey problem... what is the problem to be solved exactly? The way it is set up today, I think it will either use no proxy at all, or use the default proxy that is configured in the maven configuration. (Given you have a proxy that lets you download stuff from maven central, it should be possible to make it download node/npm as well)

bradfordwagner:
In certain enterprise solutions the settings.xml is not modifiable as it is controlled by another jenkins user which normally does not have proxy configurations. In our case we use our own in house artifactory to resolve artifacts not maven central. Adding proxy settings to the settings.xml for the build user isn't realistic in our situation. Agreed it is an edge case, but I can't have jenkins builds in house without this change.

eirslett:
There is another option specifically for this use case; `<nodeDownloadRoot>` and `<npmDownloadRoot>`; people upload the node/npm packages to their private artifactories and configure the plugin to download from there.

Flasssh (approved):
This change would be great.

Issue #637, opened 2017-08-04
install-node-and-npm : option to download a list of OS


Rouche:
Hi

**Do you want to request a _feature_ or report a _bug_?**
Probably a feature, i did not find an options like this.

**What is the current behavior?**
We package a pre build node with lots of files to speedup out build processes.
Problem is we build on linux and windows. We need both versions in the package but the mojo download the node version for the OS it run in.

**What is the expected behavior?**
Could it be possible to add an option with a list of OS for install-node-and-npm?
Meanwhile I will try to figure it out in a fork.

**Please mention your frontend-maven-plugin and operating system version.**
Linux / Windows 10

Thanks.

eirslett:
This sounds like a niche use case, I think we can consider it if many people ask for it, but for now it would probably be a lot of complexity in the plugin for a feature that would rarely be used.
I suggest you try to use exec-maven-plugin instead, try looking at this guide: 
https://addyosmani.com/blog/making-maven-grunt/

Rouche:
Actually, its not that bad. Im almost done. Ill send you a PR if you want.

We had checked ways to use exec-maven but with lots of drawbacks on build times and disk spaces. Its been a long time tough.

eirslett:
Another suggestion; just git commit the node.js binaries for windows and linux directly into the repository. It's a good complexity/efficiency trade-off for some edge cases.

You could eventually try maven-download-plugin to download some of the binaries, if you replicate the directory structure.

PR #639, opened 2017-08-10
Feature for #637

Rouche:
**Summary**

Hi,

I said i was working on it so here we go.
Its a very specific case where we prepackage a node distribution for our builds so no download occur and the dependencies are all setup already.

Right now the package is built on Linux but executed locally on windows. And in this case, the windows binary is not there.

This adds the option to download multiple node distribution to be used at a later time. Right now works with indtall-node-and-npm only.

As you wish if you want to merge or not. No regressions done.


Rouche:
This will never work. Im stuck with the same problem but with node-sass.

Issue #661, opened 2017-10-23
Refine a path resolution for NodeTaskExecutor (gulp related issue)


rzymek01:
**Do you want to request a _feature_ or report a _bug_?**

I'd like to suggest less error-prone solution for #649 (related to "cannot find module `gulp`").

**What is the current behavior?**

See #649 and #658.

Basically, both `gulp` and `gulp.js` are valid executables from node point of view, so `frontend-maven-plugin` may support both, making it resilient to extension changes of particular scripts such as `gulp`.
Looking at git history of GulpRunner “js” extension is going back and forth, which seems to support the need of the stable solution.

**If the current behavior is a bug, please provide the steps to reproduce.**

See #649 and #658.

**What is the expected behavior?**

- `gulp` executable is found regardless of extension used (supported by node, i.e. no extension, `js`, `json`, and `node`,
- correct log message should be displayed when trying to run a script from installation directory instead of working directory,
- error message should be more precise, e.g.;
```[ERROR] Failed to execute goal com.github.eirslett:frontend-maven-plugin:1.7-SNAPSHOT:gulp (gulp process-resources) on project my-project: Failed to run task: 'gulp process-resources --skip-minification=false --production=false' failed. Could not locate a task <node_modules/gulp/bin/bad-path/gulp> in neither working dir </Users/my-project/a/b/../..> nor install dir </Users/my-project/a/b/../../target> -> [Help 1]```
- I don’t know the use case of running executable from installation dir instead of working dir. I’d get rid of this, if possible.

**Please mention your frontend-maven-plugin and operating system version.**

frontend-maven-plugin version: 1.6
Mac OS X 10.11.6

rzymek01:
Bump, is this repo maintained by someone at all?

PR #662, opened 2017-10-23
issue #661: Refine a path resolution for node tasks

rzymek01:
**Summary**

- an executable (such as gulp) is found regardless of extension used (supported by node, i.e. no extension, js, json, and node)
- revamp log messages
- reuse paths-related utils

**Tests and Documentation**

I've tested the change locally in a few scenarios with different correct/invalid task paths.

I can see `gulp` happy path is already covered in `it/example project`. I'm wondering why it didn't fail for `1.6` release... Should I add new directory in `it` especially for this change?
IMO docs change it's not needed here, because it's about internal behaviour, but I might be wrong.


rzymek01:
@eirslett bumping

Issue #667, opened 2017-11-08
node-gyp is not executable


scottfwalter:
After running mvn com.github.eirslett:frontend-maven-plugin:install-node-and-npm

I noticed that node-gyp is not executable on both macOS and Linux using 1.6

❯ ls -l node/node_modules/npm/bin/node-gyp-bin
total 8
-rw-r--r-- 1 scott staff 172 Nov  8 15:07 node-gyp
-rw-r--r-- 1 scott staff 144 Nov  8 15:07 node-gyp.cmd

PR #668, opened 2017-11-10
Modified the NodeInstaller to use a custom method for copying files c…

marcajimenez:
…opyDirectoryContents() rather than using the FileUtils.copyDirectory() which does not preserve file attributes. Also modified the ArchiveExtractor class to use a umask of 0001 when checking to set execute permission on files being unpacked.

**Summary**

We recently upgraded the version of Node for our webapp application to 6.11.3 and when using the frontend-maven-plugin to help with installation we ran into a permission error at build time.  Several scripts included in the Node 6.11.3 archive which should have execute permission were not longer set to executable after the plugin's lifecycle.  This fix is meant to cover this issue posted: https://github.com/eirslett/frontend-maven-plugin/issues/667

The root cause of the issue was in the plugin's use of FileUtils.copyDirectory which does not preserve File attributes.  To fix this issue I have implemented a new method to copyDirectory contents which is used instead of the FileUtils method.

We have tested the fix on our local developer machines and jenkins machine and have confirmed that our Node installation has all of the correct permissions now.

### DIFF
--- a/frontend-plugin-core/pom.xml
+++ b/frontend-plugin-core/pom.xml
@@ -58,5 +58,140 @@
             <artifactId>slf4j-api</artifactId>
             <version>1.7.5</version>
         </dependency>
+        <dependency>
+		    <groupId>junit</groupId>
+		    <artifactId>junit</artifactId>
+		    <version>4.12</version>
+		</dependency>
     </dependencies>
+    
+	<build>
+		<pluginManagement>
+		    <plugins>
+		    		<plugin>
+			    		<artifactId>maven-antrun-plugin</artifactId>
+			    		<version>1.8</version>
+					<executions>
+						<execution>
+							<phase>integration-test</phase>
+							<configuration>
+								<target>
+								
+								<mkdir dir="${project.build.directory}/copy-src/parent"/>
+								<mkdir dir="${project.build.directory}/copy-src/parent/child"/>
+								<mkdir dir="${project.build.directory}/copy-src/parent/child/grandchild"/>
+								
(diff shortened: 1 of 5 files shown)


eirslett:
Is it possible to do this with the native Java API? https://stackoverflow.com/questions/6838221/how-to-preserve-file-permissions-when-using-fileutils-copydirectory
I haven't looked much into it.

marcajimenez:
I came across the same article and this is what I used to implement the copyDirectory function. The limitation of Files.copy() api is that it only copies a target file or directory but not the contents of a directory.

https://docs.oracle.com/javase/tutorial/essential/io/copy.html

My copyDirectory function will use Files.copy as it recurses the directory and sub-directories copying all contents.

eirslett:
Ah ok, I see.
1) Maybe move that function into its own file, as a utility class with a static method in it?
2) Is it possible to shoehorn a test case into one of the integration test cases, to verify that permissions etc are correct?
3) Looks like the Travis build is failing :-/

marcajimenez:
Shouldn't be a problem to add a test, and I'll move it to it's own file.  I should be able to get to that early this week.

thombaynes:
I've also run into this issue. I am unable to run jest tests with Node 8.9.3 and npm 5.5.1.

`[INFO] > spark_tests@1.0.0 basicTests /Users/thombaynes/Projects/Development/design/target/test-classes
[INFO] > jest --config=./config/basicConfig.json --forceExit; exit 0
[INFO]
[ERROR] sh: /Users/thombaynes/Projects/Development/design/target/test-classes/node_modules/.bin/jest: Permission denied`

rjimgal:
Is there anything preventing this PR to be merged? I've come across this issue as well, and I'm happy to help if needed.

PR #672, opened 2017-11-24
fix the yarn 1.x proxy issue

lmbai:
**Summary**

From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.

Under yarn 1.1.0, if we run some command like:
‘yarn webpack:main:build ‘ this dependence may add some extra parameters according to the system settings 
It may become to ‘ yarn webpack:main:build —http-proxy://xxxxxxxx’
It will fail because of the ‘—‘

**Tests and Documentation**

check the current yarn version. if above 1. : remove the `--` for the parameter list


lmbai:
mis understand

PR #714, opened 2018-04-29
Update tests and 3rd party libs in tests

pitgrap:
**Summary**

This will update all dependencies to the latest versions.
- use nodejs latest lts for tests
- update  all other libs to avoid a lot of warnings
- force tls 1.2 for ssl downloads, since github dropped weak cryptographic standards
- remove yarn 0.16 and make the latest version the default

**Tests and Documentation**

this should fix appveyor (which still uses JDK 1.7)


eirslett:
LGTM, thanks!

PR #737, opened 2018-06-26
Added NodeMojo for low level call to node

treilhes:
I intend to use it to call pnpm cli. So instead of implementing another specific cli i created a more generic approach.
I hope you will find it usefull.
Feel free to review/comment/request changes.


treilhes:
Sorry but i don't know how to fix build failures.If you can explain wha t i'm doing wrong i would be glad to solve it.

PR #739, opened 2018-07-06
fix(arguments): respect quotes splitting arguments by whitespaces (#336)

### LINKED ISSUE #336: Wrong parsing of arguments in install-node-and-nmp
I need to provide custom `.npmrc` to the `npm install` command. The `.npmrc` is located in root of my project. My plugin configuration looks as:

```
<!--Install dependencies using npm-->
<execution>
    <id>npm install</id>
    <goals>
        <goal>npm</goal>
     </goals>
     <configuration>
         <arguments>install --userconfig "${basedir}/.npmrc"</arguments>
     </configuration>
 </execution>
```

The quotes `""` are provided because this runs on my Jenkins CI server and it constructs the temp folder for CI build out of project name which contains spaces. Unfortunately the the args are simply split on `\s+` which ignores the quotes hence cannot be used with custom `.npmrc` located in path which contains spaces. The issue can be seen here:

https://github.com/eirslett/frontend-maven-plugin/blob/master/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NodeTaskExecutor.java#L79

IMHO the problem is you take all the arguments as single String. I suggest you have option to provide `List` of arguments instead and take them as they are. To make the change backward compatible you can provide new list property like `<argumentsList>` and use the one which is set, fail if both are set. The fixed `<configuration />` may look like:

```
<!--Install dependencies using npm-->
<execution>
    <id>npm install</id>
    <goals>
        <goal>npm</goal>
     </goals>
     <configuration>
         <argumentsList>
            <argument>install</argument>
            <argument>--userconfig</argument>
            <argument>"${basedir}/.npmrc"</argument>
         <argumentsList>
     </configuration>
 </execution>
```

Overall doing split on space looks dangerous to me. Let me know if you need help with this, will be glad to assist.

Buuhuu:
This PR improves the handling of quotes in arguments passed in through execution configurations. It modifies the behaviour of splitting the given arguments line by whitespace, by ignoring whitespaces that are enclosed in double quotes, single quotes or either of the both and the end of the line. 

A unit test for the new ArgumentsParser is provided covering 100% of its functionality. Documentation is available as Javadoc of the ArgumentsParser#parse(String) method.

Fixes #336

### DIFF
--- a/frontend-plugin-core/pom.xml
+++ b/frontend-plugin-core/pom.xml
@@ -58,5 +58,12 @@
             <artifactId>slf4j-api</artifactId>
             <version>1.7.5</version>
         </dependency>
+
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.12</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ArgumentsParser.java
+++ b/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ArgumentsParser.java
@@ -7,11 +7,6 @@ import java.util.List;
 
 class ArgumentsParser {
 
-    enum State {
-        NEUTRAL,
-        QUOTED
-    }
-
     private final List<String> additionalArguments;
 
     ArgumentsParser() {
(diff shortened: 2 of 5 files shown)


Buuhuu:
@eirslett any update on that?

eirslett:
I guess I can merge this. But try not to put to complicated stuff inside `<arguments>`, there are probably better ways to solve the problem at hand.

PR #749, opened 2018-07-26
Fix directory path too long problem on windows

edwardzjl:
**Summary**

Current version has some problems when installing npm on windows, because the installation will download a bunch of node modules in tmp dir and the path would look like "tmp/node/node_modules/module_a/node_modules/module_b/............."
But Windows cannot deal with paths longer than 247 chars.

This commit checks the path's length and will flattern any long path folder in the tmp dir, before they are deleted.

**Tests and Documentation**

Fix directory path too long problem on windows.


Zardoz89:
@edwardzjl Could do you do a rebase, so appveyor and travis can redo the tests ?

edwardzjl:
@Zardoz89 I dont know whats wrong, I can build successfully on my computer.
the appveyor log says  could not transfer artifact org.sonatype.oss:oss-parent:pom:7 due to access denied

Zardoz89:
Appveyour fails randomly. I tried to fix it, but sometime keeps failing. I don't have clue why.

On this case, it worked with JDK 1.8 on Windows, and on Travis. So I would say to go with this, but @eirslett is who decide.

Santoshbhad:
Thank you so much this helps me to fix this issue  i need to just run below command
npm cache clean --force
mvn clean install

eirslett:
Looks like a lot of unrelated changes in the same PR... could you please rebase and squash the commits so it's easier to review?

paul-bjorkstrand:
This looks like a problem with Java/windows, and how Java needs to delete folders (deleting contents before deleting the folder, recursively). I think you just need to catch the error, and log out message saying that "the temp dir was not able to be deleted, please delete it manually: <tempdir path>".

PR #761, opened 2018-08-13
Support NPM Registry authentication with credentials from Maven settings

dinix2008:
Our company uses a private proxy repository (Nexus) to access the central NPM and Maven repository and our private packages (npm, maven). The task was set for developers to use the general scenario of access settings to repositories from Maven settings.

I added the configuration parameter **npmRegistryServerId** for npm and yarn goal. Thus, you can use credentials from Maven settings for authentication in NPM Registry.

An example of the settings can be found in integration test projects (frontend-maven-plugin/src/it)

### DIFF
--- a/frontend-maven-plugin/pom.xml
+++ b/frontend-maven-plugin/pom.xml
@@ -53,6 +53,13 @@
       <artifactId>plexus-build-api</artifactId>
       <version>0.0.7</version>
     </dependency>
+
+    <dependency>
+      <groupId>org.mock-server</groupId>
+      <artifactId>mockserver-client-java</artifactId>
+      <version>5.3.0</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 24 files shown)


eirslett:
I decided not to support this through the Maven plugin, in order to reduce the "surface area" between Maven and npm. The point is to make your frontend build more independent from Maven. Some developers may want to use Gradle instead of Maven, but with the same internal npm registry. (Or even use node.js directly)
You can use [npmrc](https://docs.npmjs.com/files/npmrc) to configure servers "the npm way"; that will also work across projects.

dinix2008:
Your position is understandable. Our situation is that part of the development frontend is basically a java project and most of the team members are java developers, it was for them that the revision was done so that they set up only settings.xml once, without going into the knowledge of the npm configuration.

froblesmartin:
It is sad to not have this feature on the plugin 😞

PR #786, opened 2019-01-06
fix 'npm not found' during yarn execution

thomschke:
**Summary**

Yarn execution fails with error message `com.github.eirslett.maven.plugins.frontend.lib.DefaultYarnRunner - /bin/sh: 1: npm: not found`


thomschke:
I changed all scripts in `package.json ` to use yarn (instead npm), so this problem doesn't exist anymore.

Issue #796, opened 2019-02-14
Maven proxy settings are incorrectly passed to yarn


Borewit:
Version 1.6

frontend-maven-plugin calls yarn with:
```
yarn run build-prod --https-proxy=http://proxy:8080 --proxy=proxy:8080
```

Causing:
```
[ERROR] Unknown option: '--https-proxy'
[ERROR] Unknown option: '--proxy'
```

I think it should be:
```
yarn --https-proxy=http://proxy:8080 --proxy=proxy:8080 run build-prod
```

OldIMP:
same bug for npm

phuongnq:
Do we have any update on this issue?

mshepel:
It's resolved in version 1.12.0

neloman3:
There is the same problem with npm but 1.12.0 seems to be already bad

OmarHawk:
> It's resolved in version 1.12.0

It is not - and the PR is also still open. @eirslett any means to fix this issue from the plugin side?

gstolz:
problem also exists in version 1.12.1. Would be great to have a solution for this.
![Screenshot 2022-03-07 at 16 51 26](https://user-images.githubusercontent.com/10153918/157068831-05c13340-b7c6-4ccd-a272-e8865724190d.png)

PR #797, opened 2019-02-14
Fix yarn execution with Maven proxy defined

Borewit:
Issue #796


Borewit:
Maybe the arguments should be split it two:

1. Arguments passed to yarn, e.g. proxy
2. Arguments passed to script

If the second one is required, please let me know.

Issue #794, opened 2019-02-11
Could not extract archive: '/xxx/xyz/.m2/repository/com/github/eirslett/node/11.9.0/node-11.9.0-linux-x64.tar.gz': EOFException


Santoshbhad:
Failed to execute goal com.github.eirslett:frontend-maven-plugin:1.6:install-node-and-npm (default) on project : Could not extract the Node archive: Could not extract archive: '/xxx/xyz/.m2/repository/com/github/eirslett/node/11.9.0/node-11.9.0-linux-x64.tar.gz': EOFException 

Please help me understand and fix this issue.
Thanks

Zardoz89:
See issue #738 . PR https://github.com/eirslett/frontend-maven-plugin/pull/749 should fix it

seregamorph:
@santoshbhad @zardoz89 @eirslett I had the same issue locally and investigated it.
Please take a look https://github.com/eirslett/frontend-maven-plugin/pull/807 with suggested fix.

seregamorph:
This issue can be closed (addressed in #807)

PR #807, opened 2019-04-16
#794 Node, npm and yarn installer delete corrupted archive

seregamorph:
**Summary**
Issue #794
The node, npm and yarn installer deletes corrupted archive and destination directory in case if the downloading was interrupted (causes EOFException when extracted).

**Detailed**
The possible real-life example:
* The build is started and at the phase of downloading npm, node or yarn it is interrupted on the slow connection, the build fails
* Re-start the build - it has lazy download logic and determines that the destination archive is already in place
* The archive is extracted and fails with root cause EOFException, the build fails
  * Re-start the build and either node (npm / yarn file) is here, but the package is incomplete - the build may fail
  * or the previous step (extract archive) is repeated again and again fails.
This requires manual investigation and deleting the cached archive and can take much time.

How to simulate it locally.
* Make a successful build of npm (yarn) project.
* Go to `$HOME/.m2/repository/com/github/eirslett/node/${node.version}` directory
* Rename the existing archive to `-full` suffix and make it's partial copy of first bytes:
```
dd if=node-8.9.4-darwin-x64.tar.gz-full of=node-8.9.9-darwin-x64.tag.gz bs=1000000 count=1
```
* Remove project cached node version
* Run the build
* Re-run the build

Please note, that no integration tests are provided for the reasons of non-clear approach for it (the build should be interrupted while the archive is being downloaded; or it can be pre-cached manually to the maven directory). I tested it locally step-by-step.

**Also please note, that there is another way to solve this:**
The downloading should be done to a temp file unless completed and then renamed to destination. Both approaches can be mixed in case if someone will face the described above scenario expecting that it will be solved with plugin update.

### DIFF
--- a/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NPMInstaller.java
+++ b/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NPMInstaller.java
@@ -1,5 +1,6 @@
 package com.github.eirslett.maven.plugins.frontend.lib;
 
+import java.io.EOFException;
 import java.io.File;
 import java.io.IOException;
 import java.util.Arrays;
@@ -139,12 +140,28 @@ public class NPMInstaller {
                 this.logger.warn("Failed to delete existing NPM installation.");
             }
 
-            extractFile(archive, nodeModulesDirectory);
+            File packageDirectory = new File(nodeModulesDirectory, "package");
+            try {
+                extractFile(archive, nodeModulesDirectory);
+            } catch (ArchiveExtractionException e) {
+                if (e.getCause() instanceof EOFException) {
+                    // https://github.com/eirslett/frontend-maven-plugin/issues/794
+                    // The downloading was probably interrupted and archive file is incomplete:
+                    // delete it to retry from scratch
+                    this.logger.error("The archive file {} is corrupted and will be deleted. "
+                            + "Please try the build again.", archive.getPath());
+                    archive.delete();
+                    if (packageDirectory.exists()) {
+                        FileUtils.deleteDirectory(packageDirectory);
+                    }
+                }
+
+                throw e;
+            }
 
             // handles difference between old and new download root (nodejs.org/dist/npm and
             // registry.npmjs.org)
             // see https://github.com/eirslett/frontend-maven-plugin/issues/65#issuecomment-52024254
-            File packageDirectory = new File(nodeModulesDirectory, "package");
             if (packageDirectory.exists() && !npmDirectory.exists()) {
                 if (!packageDirectory.renameTo(npmDirectory)) {
                     this.logger.warn("Cannot rename NPM directory, making a copy.");
--- a/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NodeInstaller.java
+++ b/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NodeInstaller.java
@@ -1,5 +1,6 @@
 package com.github.eirslett.maven.plugins.frontend.lib;
 
+import java.io.EOFException;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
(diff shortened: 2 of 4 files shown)


frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NPMInstaller.java
```diff
@@ -139,12 +140,28 @@ private void installNpm() throws InstallationException {
...
+                extractFile(archive, nodeModulesDirectory);
+            } catch (ArchiveExtractionException e) {
+                if (e.getCause() instanceof EOFException) {
+                    // https://github.com/eirslett/frontend-maven-plugin/issues/794
+                    // The downloading was probably interrupted and archive file is incomplete:
+                    // delete it to retry from scratch
+                    this.logger.error("The archive file {} is corrupted and will be deleted. "
+                            + "Please try the build again.", archive.getPath());
+                    archive.delete();
+                    if (packageDirectory.exists()) {
```
seregamorph: the "package" directory is partially filled (dirty), for this reason should be purged

frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NodeInstaller.java
```diff
@@ -140,7 +141,21 @@ private void installNodeDefault() throws InstallationException {
...
+                extractFile(archive, tmpDirectory);
+            } catch (ArchiveExtractionException e) {
+                if (e.getCause() instanceof EOFException) {
+                    // https://github.com/eirslett/frontend-maven-plugin/issues/794
+                    // The downloading was probably interrupted and archive file is incomplete:
+                    // delete it to retry from scratch
+                    this.logger.error("The archive file {} is corrupted and will be deleted. "
+                            + "Please try the build again.", archive.getPath());
+                    archive.delete();
+                    FileUtils.deleteDirectory(tmpDirectory);
```
seregamorph: no need to check that `tmpDirecory` exists (guaranteed by the flow)

frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/YarnInstaller.java
```diff
@@ -122,7 +123,23 @@ private void installYarn() throws InstallationException {
...
+                extractFile(archive, installDirectory);
+            } catch (ArchiveExtractionException e) {
+                if (e.getCause() instanceof EOFException) {
+                    // https://github.com/eirslett/frontend-maven-plugin/issues/794
+                    // The downloading was probably interrupted and archive file is incomplete:
+                    // delete it to retry from scratch
+                    this.logger.error("The archive file {} is corrupted and will be deleted. "
+                            + "Please try the build again.", archive.getPath());
+                    archive.delete();
+                    if (installDirectory.exists()) {
```
seregamorph: The above code cleanups the installDirectory, but there is a probabilty to use corrupted (incomplete) Yarn dist.

frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NPMInstaller.java
```diff
@@ -139,12 +140,28 @@ private void installNpm() throws InstallationException {
...
+                        FileUtils.deleteDirectory(packageDirectory);
+                    }
+                }
+
+                throw e;
```
seregamorph: the possible alternate option here is to re-download one more time
but the probability of this failure is minor, so in case if it happens the user should not even understand what happened :D

eirslett:
Thanks, that looks good!

frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NPMInstaller.java
ryanrupp: Related issue - https://github.com/eirslett/frontend-maven-plugin/issues/882 - auto-delete triggers but the auto-retry probably would be nice as a future enhancement

ryanrupp: Actually, downloader would be benefit from downloading to a tmp location and then moving post download, I put a comment [here](https://github.com/eirslett/frontend-maven-plugin/issues/882#issuecomment-1542390591).

PR #809, opened 2019-04-25
Added support for Raspbian OS armv7l architecture

codelano:
Currently, when using install-node-and-npm on Rasbian OS, it resolves it as x86 arch, which is not correct and it's unable to donwload it from https://nodejs.org/dist/{node.version}/node-{node.version}-linux-{os.arch}.tar.gz.
This change, allows for Rasbian env to resolve proper archive name.

Compiled and tested on Rasbian with RPI 3B+


eirslett:
thanks! I have no way of testing this automatically in CI, so let's hope it won't break.

tchepro:
Is this change going to be merged to master? It would be nice with PI support.

eirslett:
Yes it should! I'll get it in now. Release when I have time.

eirslett:

PR #810, opened 2019-05-07
fix parsing repeating arguments

jpodeszwik:
**Summary**

Sometimes it might be necessary to pass arguments with repeating value, for example building Angular application you can use:

```npm run build -- --base-href /some/href/ --deploy-url /some/href/```

which is valid build command, but this plugin tries to remove duplicates, so resulting command will be:

```npm run build -- --base-href /some-href/ --deploy-url```.

There are workarounds possible, but it still may be confusing for users if some args are removed.

This PR changes behavior, so that it's now possible to add duplicates into the build command. Additional parameters are still checked for duplicates.

**Tests and Documentation**

Added changelog info.
Added unittest.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Last public release: [![Maven Central](https://maven-badges.herokuapp.com/maven-
 
 ## Changelog
 
+### 1.7.7
+
+* Allow repeating arguments
+
 ### 1.7.6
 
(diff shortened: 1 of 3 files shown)


jpodeszwik:
AppVeyor build fails on javadoc generation due to bug in JDK:
https://bugs.openjdk.java.net/browse/JDK-8212233

jpodeszwik:
Closing this as similar change from #930 is merged

PR #811, opened 2019-05-09
Added support to disable proxy when downloading node,npm and yarn

Francis-FY:
Added `installNodeAndYarnDirectly` and `installNodeAndNpmDirectly` property

**Summary**

Add this two property to disable maven proxy when downloading node, npm and yarn. 

**Tests and Documentation**

The plugin uses the maven proxy settings to download node,npm and yarn by default, it may cause the download to fail if the resource cannot be accessed through the proxy, and there is no way to edit the maven settings file to change the proxy setting, you can configure the plugin like this to download node,npm and yarn directly instead of using maven proxy:

```xml
<plugin>
    ...
    <configuration>
        <installNodeAndNpmDirectly>false</installNodeAndNpmDirectly>
        <installNodeAndYarnDirectly>false</installNodeAndYarnDirectly>
    </configuration>
</plugin>

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -443,6 +443,22 @@ If you want to disable proxy for Yarn you can use `yarnInheritsProxyConfigFromMa
 
 ```
 
+If you want to disable proxy for installing node, npm and yarn you can use `installNodeAndNpmDirectly`(for npm)
+or `installNodeAndYarnDirectly`(for yarn). The plugin uses the maven proxy settings in your settings.xml file to
+download node,npm and yarn by default, it may cause the download to fail if the resource cannot be accessed through
+your proxy,you can configure the plugin like this to download node,npm and yarn directly:
+
+```xml
+<plugin>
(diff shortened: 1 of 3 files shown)


starkipraggy:
Have you tried setting nonProxyHosts on your maven proxy config?
If that works I don't really see the need for this PR.

```
<proxies>
   <proxy>
        <id>BaeldungProxy_Authenticated</id>
        <active>true</active>
        <protocol>http</protocol>
        <host>proxy.baeldung.com</host>
        <port>80</port>
        <username>baeldung</username>
        <password>changeme</password>
       <!-- nodejs.org/yarn download host here -->
        <nonProxyHosts>internal.baeldung.com|localhost|127.*|[::1]</nonProxyHosts>
    </proxy>
</proxies>
```

More details:
https://www.baeldung.com/maven-behind-proxy

PR #816, opened 2019-05-17
Fix regression when extracting tar archive

arminha:
**Summary**

Commit 93d77ffc023effbcb36813648b578a0541709d76 introduced a check that files extracted from a tar archive will not be written outside of the destination directory. Unfortunately it also introduced a regression prventing the extraction of any tar archive when the path to the destination directory contains a symbolic link.

For example on a jenkins agent where `/var/lib/jenkins` is a symbolic link to `/data/jenkins` and a job tries to extact nodejs to `/var/lib/jenkins/workspace/example-project/target/node/tmp`. The old code would then check if canonical path to a tar entry like `/data/jenkins/workspace/example-project/target/node/tmp/XXX` starts with `/var/lib/jenkins/workspace/example-project/target/node/tmp` which always fails.

The new check compares the **canonical** extraction paths of the tar entries with the **canonical** path of the destination directory. This also works when the destination directory path contains a symbolic link and still protects against extracting files outside the destination directory.

### DIFF
--- a/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ArchiveExtractor.java
+++ b/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ArchiveExtractor.java
@@ -81,53 +81,53 @@ final class DefaultArchiveExtractor implements ArchiveExtractor {
                         final File destPath = new File(destinationDirectory + File.separator + entry.getName());
                         prepDestination(destPath, entry.isDirectory());
                         if(!entry.isDirectory()){
-		                        InputStream in = null;
-		                        OutputStream out = null;
-		                        try {
-		                            in = zipFile.getInputStream(entry);
-		                            out = new FileOutputStream(destPath);
-		                            IOUtils.copy(in, out);
-		                        } finally {
-		                            IOUtils.closeQuietly(in);
-		                            IOUtils.closeQuietly(out);
-		                        }
+                            InputStream in = null;
+                            OutputStream out = null;
+                            try {
+                                in = zipFile.getInputStream(entry);
+                                out = new FileOutputStream(destPath);
+                                IOUtils.copy(in, out);
+                            } finally {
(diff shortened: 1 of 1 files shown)


DiabolusExMachina:
Could this please get merged? I cannot build my projects anymore this is kind of really bad...

eirslett:
Ok, looks good.

frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ArchiveExtractor.java
```diff
@@ -104,13 +104,14 @@ public void extract(String archive, String destinationDirectory) throws ArchiveE
                     tarIn = new TarArchiveInputStream(new GzipCompressorInputStream(fis));
 
                     TarArchiveEntry tarEntry = tarIn.getNextTarEntry();
+                    String canonicalDestinationDirectory = new File(destinationDirectory).getCanonicalPath();
                     while (tarEntry != null) {
                         // Create a file for this tarEntry
                         final File destPath = new File(destinationDirectory + File.separator + tarEntry.getName());
                         prepDestination(destPath, tarEntry.isDirectory());
-                        if (!destPath.getCanonicalPath().startsWith(destinationDirectory)) {
+                        if (!destPath.getCanonicalPath().startsWith(canonicalDestinationDirectory)) {
```
crisu: this fix was lost in 26 aug

crisu:
i think this fix was lost in 26 August 
https://github.com/eirslett/frontend-maven-plugin/blob/7d6b690447b44167f714cf81df8011eb90224313/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ArchiveExtractor.java#L113

arminha:
@crisu: you're right. It was indeed forgotten. Probably my fault for not writing tests :wink:

Please have a look at #864. It should fix the issue again and also adds some unit tests.

eirslett:

PR #824, opened 2019-06-19
Add unit tests for com.github.eirslett.maven.plugins.frontend.lib.Utils

petersear:
**Summary**

I've written some tests for the methods in `com.github.eirslett.maven.plugins.frontend.lib.Utils` with the help of [Diffblue Cover](https://www.diffblue.com/opensource).

Hopefully, these tests will help detect any regressions caused by future code changes. If you would find it useful to have additional tests written for this repository, I would be more than happy to look at other classes in a subsequent PR.

**Tests and Documentation**
This PR provides unit tests for existing code.

### DIFF
--- a/frontend-plugin-core/src/test/java/com/github/eirslett/maven/plugins/frontend/lib/UtilsTest.java
+++ b/frontend-plugin-core/src/test/java/com/github/eirslett/maven/plugins/frontend/lib/UtilsTest.java
@@ -1,0 +1,51 @@
+package com.github.eirslett.maven.plugins.frontend.lib;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class UtilsTest {
+
+  @Test
+  public void testImplode() {
+    final String separator = "Bar";
+    final List<String> elements = new ArrayList<String>();
+
+    assertEquals("", Utils.implode(separator, elements));
+
+    elements.add("foo");
+    elements.add("bar");
+    assertEquals("foo bar", Utils.implode(separator, elements));
+  }
+
+  @Test
+  public void testIsRelative() {
+    assertTrue(Utils.isRelative("foo/bar"));
+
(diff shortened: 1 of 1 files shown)


frontend-plugin-core/src/test/java/com/github/eirslett/maven/plugins/frontend/lib/UtilsTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+    assertFalse(Utils.isRelative("/foo/bar"));
+    assertFalse(Utils.isRelative("file:foo/bar"));
+    assertFalse(Utils.isRelative("C:\\SYSTEM"));
+  }
+
+  @Test
+  public void testMerge() {
+    final List<String> first = new ArrayList<String>();
+    first.add(null);
```
eirslett: why is it adding `null`?

petersear: I've changed these inputs and those to the testPrepend() method to valid strings and removed the `null`.

frontend-plugin-core/src/test/java/com/github/eirslett/maven/plugins/frontend/lib/UtilsTest.java
```diff
@@ -0,0 +1,63 @@
...
+    first.add("foo");
+    final List<String> second = new ArrayList<String>();
+    second.add("bar");
+    final List<String> actual = Utils.merge(first, second);
+
+    final List<String> arrayList = new ArrayList<String>();
+    arrayList.add("foo");
+    arrayList.add("bar");
+
+    assertEquals(arrayList, actual);
```
eirslett: I guess it's technically correct, but maybe this would be easier to read:
```
assertEquals(
  Arrays.asList("foo", "bar"),
  Utils.merge(Arrays.asList("foo"), Arrays.asList("bar")
)
```
or what do you think?

petersear: Yes, I agree. I am all in favour of increasing readability!
I have changed as suggested and similarly for `testPrepend()`.

eirslett:
ok thanks!

PR #828, opened 2019-07-04
Fix link to Sonatype

Blacktiger:
**Summary**

It appears that the old link in the docs pointed to an NXRM 2 page, except it was trying to get to an NXRM 3 page. I updated the link.

**Tests and Documentation**

N/A


eirslett:
Thanks!

PR #832, opened 2019-07-19
fix

StNekroman:
Add support for multiple PATH variables on Windows


eirslett:
Ok thanks!

StNekroman:
When this will be published to mvn repository?

PR #834, opened 2019-08-06
Always log to stdout

### LINKED ISSUE #833: stderr output does not imply error
Looking at https://github.com/webpack/webpack-cli/issues/1017#issuecomment-517569237 and the various examples mentioned at https://github.com/eirslett/frontend-maven-plugin/issues/584 it's pretty clear that flagging all stderr output as `[ERROR]` is a but, not a feature.

Please add an option to disable this feature... ideally by default. At the very least we should be able to disable it if we want to.

Secondly, is there any chance of us getting colored output? Maven 3.5.0+ outputs ANSI colors. Yarn outputs ANSI colors. But when we run Yarn through this plugin we lose all colors. Instead of redirecting output to the log, would it help to output to stdout, stderr directly instead?

eirslett:
This should fix #833 .
@cowwoc could you please try to build this branch, and then build your project with the snapshot version of the plugin - does that work as intended?


cowwoc:
@eirslett Confirmed. This works as expected.

Issue #813, opened 2019-05-14
Could not extract the Node archive on OSX


otrebski:
**Do you want to request a _feature_ or report a _bug_?**
Bug

**What is the current behavior?**
Fails to download node on OSX with following error:
```
[ERROR] Failed to execute goal com.github.eirslett:frontend-maven-plugin:1.7.6:install-node-and-npm (install node and npm) on project *********: Could not extract the Node archive: Could not extract archive: '/Users/*****/.m2/repository/com/github/eirslett/node/10.15.3/node-10.15.3-darwin-x64.tar.gz': Expanding node-v10.15.3-darwin-x64/ would create file outside of /tmp/node/tmp -> [Help 1]
```
**If the current behavior is a bug, please provide the steps to reproduce.**
On OSX run `mvn install`

**What is the expected behavior?**
npm will be installed and run build

**Please mention your frontend-maven-plugin and operating system version.**
Plugin version: frontend-maven-plugin:1.7.6
OS: OSX 18.5.0 Darwin Kernel Version 18.5.0

margingrotle:
I am experiencing the same.

MikeThomsen:
@eirslett I just sent you a PR that fixes this on my MBP. The issue that I found was that it's triggered on macOS by a case-insensitive path. For some reason, a folder in my repo that is named `Frontend` was getting processed as `frontend` by this plugin (or vice versa, regardless it was case sensitivity).

I added new checks for case sensitive file systems to `ArchiveExtractor` and was able to build the project just fine using that fork of the plugin. PR is [here](https://github.com/eirslett/frontend-maven-plugin/pull/843)

aozolin:
I have exactly the same issue on Windows 10 OS.
It works fine with 

> frontend-maven-plugin:1.6
But doesn't work with 1.7.x and 1.8.0
Could you fix it please?

`Failed to execute goal com.github.eirslett:frontend-maven-plugin:1.7.5:install-node-and-npm (+++ install node and npm into core +++) on project test-project: Could not extract the npm archive: Could not extract archive: 'C:\Users\Administrator\.m2\repository\com\github\eirslett\npm\6.9.0\npm-6.9.0.tar.gz': Expanding package/package.json would create file outside of c:\Jenkins\workspace\Test Project\test-project\src\main\frontend\src\test\node\node_modules`

MikeThomsen:
[...]

PR #843, opened 2019-08-26
Fixed a bug in ArchiveExtractor that caused it to fail on case-insens…

MikeThomsen:
…itive file systems.

**Summary**

**Tests and Documentation**

### DIFF
--- a/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ArchiveExtractor.java
+++ b/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ArchiveExtractor.java
@@ -46,7 +46,7 @@ final class DefaultArchiveExtractor implements ArchiveExtractor {
             if (!path.getParentFile().exists()) {
                 path.getParentFile().mkdirs();
             }
-            if(!path.getParentFile().canWrite()) {
+            if (!path.getParentFile().canWrite()) {
                 throw new AccessDeniedException(
                         String.format("Could not get write permissions for '%s'", path.getParentFile().getAbsolutePath()));
             }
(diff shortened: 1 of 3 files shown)


eirslett:
Thanks! Looks like this patch is needed by more people (#813) but the tests fail on Travis... could you see if you can figure out the reason why, and make them pass somehow?

MikeThomsen:
@eirslett I'll try to find some time soon to figure out what's going on.

MikeThomsen:
It was an `assert` in `yarn-integration/verify.groovy` that was causing it. The relevant line it was looking for no longer appears at all in the build log. Left a reminder TODO for you to see if that part of the test needs something new.

frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ArchiveExtractor.java
```diff
@@ -139,4 +141,46 @@ public void extract(String archive, String destinationDirectory) throws ArchiveE
...
+            /*
+             * The first check should handle case-sensitive file systems. We need this
+             * in order to weed out case-sensitive file systems with a non-existent path
+             * that slipped through the first test.
+             */
+            if (new File(destPath).exists() && !(new File(destPath.toLowerCase()).exists())) {
+                return false;
+            }
+
+            boolean retVal = true;
```
eirslett: Why is this stuff going on? Isn't it easier with `destPath.toLowerCase().startsWith(destDir.toLowerCase())`?

frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ArchiveExtractor.java
```diff
@@ -139,4 +141,46 @@ public void extract(String archive, String destinationDirectory) throws ArchiveE
...
+    /**
+     * Do multiple file system checks that should enable the plugin to work on any file system
+     * whether or not it's case sensitive or not.
+     *
+     * @param destPath
+     * @param destDir
+     * @return
+     */
+    private boolean startsWithPath(String destPath, String destDir) {
+        if (destDir.startsWith(destDir)) {
```
eirslett: Won't this check always be true? Is it supposed to be `destPath`?

MikeThomsen: That was a typo on my part.

frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ArchiveExtractor.java
```diff
@@ -139,4 +141,26 @@ public void extract(String archive, String destinationDirectory) throws ArchiveE
...
+     * @param destDir
+     * @return
+     */
+    private boolean startsWithPath(String destPath, String destDir) {
+        if (destPath.startsWith(destDir)) {
+            return true;
+        } else if (destDir.length() > destPath.length()) {
+            return false;
+        } else {
+            if (new File(destPath).exists() && !(new File(destPath.toLowerCase()).exists())) {
```
eirslett: I don't understand why this check is here, why is it checking whether a file exists?

MikeThomsen: The path that Maven was feeding your plugin was not the same case as what was on the file system. However, HPFS+ and NTFS are case insensitive by default so the folder does actually exist and technically that exception should not be thrown.

FWIW, we fixed **our** issue by finally stomping the inconsistencies on a Linux box IIRC. Based on the other users' comments, I'm not entirely sure their issues are the same so you might want to just be aware of this particular scenario and ask them for more details.

MikeThomsen: That said, I think all of the other users were running Windows or macOS.

eirslett:
ok, thank you!

eirslett:

Issue #636, opened 2017-08-02
java.io.IOException: Cannot run program "/opt/app-root/src/frontend/node/node"


CanardSauvage:
**What is the current behavior?**
* Trying to run npm in a docker container with alpine linux
* the mvn install build runs normal until it hits the frontend-maven-plugin

``
[ERROR] Failed to execute goal com.github.eirslett:frontend-maven-plugin:1.5:npm (npm install) on project price-alert-backoffice-frontend: Failed to run task: 'npm install' failed. java.io.IOException: Cannot run program "/opt/app-root/src/frontend/node/node" (in directory "/opt/app-root/src/frontend"): error=2, No such file or directory
``

with

```
<plugin>
    <groupId>com.github.eirslett</groupId>
    <artifactId>frontend-maven-plugin</artifactId>
    <version>1.5</version>
    <configuration>
        <nodeVersion>v8.2.1</nodeVersion>
        <npmVerson>5.3.0</npmVerson>
    </configuration>
    <executions>
        <execution>
            <id>install node and npm</id>
            <goals>
                <goal>install-node-and-npm</goal>
            </goals>
            <phase>generate-resources</phase>
        </execution>
        <execution>
            <id>npm install</id>
            <goals>
                <goal>npm</goal>
            </goals>
            <phase>generate-resources</phase>
            <configuration>
                <arguments>install</arguments>
            </configuration>
        </execution>
        <execution>
            <id>npm build</id>
            <goals>
                <goal>npm</goal>
            </goals>
            <phase>compile</phase>
            <configuration>
                <arguments>run build</arguments>
            </configuration>
        </execution>
    </executions>
</plugin>
```

**If the current behavior is a bug, please provide the steps to reproduce.**
* created a docker image with alpine linux
* mvn 3.5.0
* create a pom with the frontend plugin

**What is the expected behavior?**
npm install works

**Please mention your frontend-maven-plugin and operating system version.**
Alpine Linux in a Docker container with frontend-maven-plugin 1.5 (also tried 1.4)

```
[INFO] Installing node version v8.2.1
--
  | [INFO] Downloading https://nodejs.org/dist/v8.2.1/node-v8.2.1-linux-x64.tar.gz to /opt/app-root/src/?/.m2/repository/com/github/eirslett/node/8.2.1/node-8.2.1-linux-x64.tar.gz
  | [INFO] No proxies configured
  | [INFO] No proxy was configured, downloading directly
[...]

Issue #633, opened 2017-07-25
node binary incompatible with Alpine, specifically Docker maven:alpine


gsharp:
**Do you want to request a _feature_ or report a _bug_?**

This is part feature request and bug... but mostly an FYI.

If you're using the Docker maven:alpine version the nodejs binary that this plugin installs will not be happy.

**What is the current behavior?**

* it will install node
* It will give an error that the node file is not found and cause a maven error

**If the current behavior is a bug, please provide the steps to reproduce.**

* it will install an incompatible binary version of node and try to run it
* it will complain that the file is not found and error out

**What is the expected behavior?**

* auto detect / install the Alpine binary (if it exists)
* provide a prerequisites info in the readme that there may be issues with the node binary not being compiled for Alpine Linux (musl)
* provide any workarounds or just say "Alpine is not supported".

**Please mention your frontend-maven-plugin and operating system version.**

* Plugin version: 1.5 
* OS: Docker maven:alpine

gsharp:
btw, the solve for this is to use Docker ```maven:latest```

stickycode:
@gsharp That is not a solution that is a workaround. 

I'm fairly confident the issue is that the default shell in alpine does not do path mapping like bash does. So executing /some/path/bin/node in a directory of /some/path tries to run /some/path/some/path/bin/node, 

If that is the case then the correct solution is to use a relative path to the executable when you provide  a working directory to the process builder.

gsharp:
@stickycode fair enough...it is indeed a workaround. I'll poke about with the paths as well, thanks for path the tip.

tholu:
Same problem here, a PR would be very nice @stickycode !

tholu:
@stickycode Do you recall in what way? We're currently experimenting with moving our Docker images for testing to Alpine and this is currently the only issue left to get everything working again. We are looking into replacing the node/yarn functionality with maven plugins right now.

tholu:
@gsharp Did you find anything regarding the path issue for a real fix? Maven plugins to compile sass and minify JS/CSS are not really working either.

stickycode:
The underlying issue I believe is the expansion of paths in the  shell .
Bash will let you define a working directory and a full path to a binary.
Whereas ash (shell in busybox) expects a relative path for the binary.

It's possible that it is an issue with the process builder. Rather that how
the plugins use it.

A workaround is to install bash in base image and invoke maven in that
shell.

flavioschuindt:
[...]

Issue #287, opened 2015-09-18
Cannot run program - node using ${basedir}/../ working directory in CI


pmafi:
Hi,

I upgraded to 0.0.25 and it no longer recognised the workingDirectory `"${basedir}/../"` correctly when i return to 0.0.23 it worked fine again.

`build	18-Sep-2015 02:25:21	[ERROR] Failed to execute goal com.github.eirslett:frontend-maven-plugin:0.0.25:karma (jasmine-integration-tests) ... java.io.IOException: Cannot run program "/opt/bamboo-agent/bamboo-agent-home/xml-data/build-dir/XXX-ZZZ-ITHSQL/project-1/project-testserver/../node/node" (in directory "/opt/bamboo-agent/bamboo-agent-home/xml-data/build-dir/XXX-ZZZ-ITHSQL/project-1/project-testserver/.."): error=2, No such file or directory -> [Help 1]`

pmafi:
In 0.0.25, i also tried configuring installDirectory as ${basedir}/../

stickycode:
I have something similar with 1.4 when running the build in docker

 Failed to execute goal com.github.eirslett:frontend-maven-plugin:1.4:npm (npm install) on project apollo-iqconnect-portal: Failed to run task: 'npm install' failed. java.io.IOException: Cannot run program "/build/node/node" (in directory "/build"): error=2, No such file or directory -> [Help 1]

rvtyagi:
Please help, I am also having the issue while running on docker

mriehema:
- Is your node/npm script working in Docker without this plugin? 
- If yes, what system platform is used in docker? Is it supported by node?

mriehema:
Looks like #633

stickycode:
I came to the same conclusion that it was #633, I'm using alpine

PR #853, opened 2019-09-27
add support for Alpine Linux using experimental nodejs download url

tnguyen1:
**Summary**

This pull request is an attempt to support _frontend-maven-plugin_ on **Alpine Linux**. The issue has been reported several times here and there: https://github.com/eirslett/frontend-maven-plugin/issues/633, https://github.com/eirslett/frontend-maven-plugin/issues/636, https://github.com/eirslett/frontend-maven-plugin/issues/287

The proposed solution is to detect when we are running under Alpine Linux (_musl libc_) using the `ldd` command. In such case, we should download from https://unofficial-builds.nodejs.org/download/release/ instead. Although experimental, these builds are stable enough for Alpine Linux in most cases.

This prevents from forcing _glibc_ into Alpine Linux, hence avoiding some annoying Warning messages.

**Caveats**

Currently the following integration test fails, but it's the same behaviour on *master* branch.
```
[ERROR] The following builds failed:
[ERROR] *  yarn-integration/pom.xml
[...]
Running post-build script: /some/path/to/frontend-maven-plugin/frontend-maven-plugin/target/it/yarn-integration/verify.groovy
java.lang.AssertionError: incorrect local repository location. Expression: buildLog.replace(java.io.File.separatorChar, (char) /).matches((?s).+Unpacking .+\Q/local-repo/com/github/eirslett/yarn/[1-9\.]*/yarn-[1-9\.]*.tar.gz\E into .+/target/node/yarn.+)
```

### DIFF
--- a/frontend-maven-plugin/src/it/custom-install-directory/pom.xml
+++ b/frontend-maven-plugin/src/it/custom-install-directory/pom.xml
@@ -27,7 +27,7 @@
                             <goal>install-node-and-npm</goal>
                         </goals>
                         <configuration>
-                            <nodeVersion>v8.11.1</nodeVersion>
+                            <nodeVersion>v8.16.1</nodeVersion>
                             <npmVersion>5.6.0</npmVersion>
                         </configuration>
                     </execution>
--- a/frontend-maven-plugin/src/it/custom-working-directory/pom.xml
+++ b/frontend-maven-plugin/src/it/custom-working-directory/pom.xml
@@ -27,7 +27,7 @@
                             <goal>install-node-and-npm</goal>
                         </goals>
                         <configuration>
-                            <nodeVersion>v8.11.1</nodeVersion>
+                            <nodeVersion>v8.16.1</nodeVersion>
                             <npmVersion>5.6.0</npmVersion>
                         </configuration>
                     </execution>
--- a/frontend-maven-plugin/src/it/example project/pom.xml
+++ b/frontend-maven-plugin/src/it/example project/pom.xml
@@ -25,7 +25,7 @@
             </goals>
             <configuration>
               <!-- See https://nodejs.org/en/download/ for latest node and npm (lts) versions -->
-              <nodeVersion>v8.11.1</nodeVersion>
+              <nodeVersion>v8.16.1</nodeVersion>
               <npmVersion>5.6.0</npmVersion>
             </configuration>
           </execution>
--- a/frontend-maven-plugin/src/it/node-provided-npm/pom.xml
+++ b/frontend-maven-plugin/src/it/node-provided-npm/pom.xml
@@ -27,7 +27,7 @@
                             <goal>install-node-and-npm</goal>
                         </goals>
                         <configuration>
-                            <nodeVersion>v8.11.1</nodeVersion>
+                            <nodeVersion>v8.16.1</nodeVersion>
                         </configuration>
                     </execution>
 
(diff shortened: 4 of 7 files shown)


digable1:
Can somebody review this please - after almost 5 months?  This is a much-desired capability that we have also just run into dur to the mixed linux distros we have that leverage _frontend-maven-plugin_

If the same test errors indeed happen in master as well as this PR, can that be a separate conversation that does not impede the review of this PR?

leungbk:
I'm on NixOS and am encountering the same issue. It would be great if this could be reviewed.

imacks:
hi i hit this issue today trying to build keycloak 10.0.1 on alpine 3.11. `node` downloaded cannot run on alpine bcos its for glibc. i had to hack the cache mechanism for it to build:

```
apk add nodejs ; \
alpine_nodever=$(node --version) ; \
alpine_nodever_num=$(echo "$alpine_nodever" | tail -c +2) ; \
kc_nodever=$(cat themes/pom.xml | grep nodeVersion | cut -d'<' -f2 | cut -d'>' -f2) ; \
sed -i "s/<nodeVersion>${kc_nodever}<\/nodeVersion>/<nodeVersion>${alpine_nodever}<\/nodeVersion>/g" themes/pom.xml ; \
mkdir -p "/tmp/node-${alpine_nodever}-linux-x64/bin" ; \
cp $(which node) "/tmp/node-${alpine_nodever}-linux-x64/bin/" ; \
tar cz -f "node-${alpine_nodever_num}-linux-x64.tar.gz" -C /tmp "node-${alpine_nodever}-linux-x64/bin/node" ; \
rm -rf "/tmp/node-${alpine_nodever}-linux-x64" ; \
mkdir -p "/root/.m2/repository/com/github/eirslett/node/${alpine_nodever_num}" ; \
mv "node-${alpine_nodever_num}-linux-x64.tar.gz" "/root/.m2/repository/com/github/eirslett/node/${alpine_nodever_num}/" ; \
```

After that `$M2_HOME/bin/mvn -Pdistribution -pl distribution/server-dist -am -Dmaven.test.skip clean install` works.

adriancole:
this would be handy for zipkin also who are trying to work in alpine

adriancole:
I would suggest that if we can control the classifier, it could be simpler.

For example, you can already override the base URL like this:

```xml
          <nodeDownloadRoot>https://unofficial-builds.nodejs.org/download/release/</nodeDownloadRoot>
```

The main thing missing is to override the classifier, which is a small part of this change.

Ex to change this
```
https://unofficial-builds.nodejs.org/download/release/v12.18.4/node-v12.18.4-linux-x64.tar.gz
```

To this:
```
https://unofficial-builds.nodejs.org/download/release/v12.18.4/node-v12.18.4-linux-x64-musl.tar.gz
```

Note there are a few classifiers as it is:
```
node-v12.18.4-linux-armv6l.tar.gz                  15-Sep-2020 23:42            21402256
node-v12.18.4-linux-x64-musl.tar.gz                15-Sep-2020 23:18            23691939
node-v12.18.4-linux-x64-usdt.tar.gz                15-Sep-2020 23:56            23013495
```

If there was a way to override `Platform.getNodeClassifier()` per property, we could fix his with environment variables instead of trying to detect. Would that be easier to merge @eirslett? or should we consider continuing this approach?

eirslett:
I guess this PR is the simplest way to implement. I'm not super happy about the complexity, but it's a feature that's requestad a lot. Looks like the build fails, though, I'm not sure why.

adriancole:
Took a different stab here cc @eirslett @anuraaga https://github.com/eirslett/frontend-maven-plugin/pull/934

eirslett:
Closed via #934 .

Issue #750, opened 2018-07-26
add support npx runner


iamruslanbakirov:
Hi, since version npm@5.2.0, a pre-installed package appeared - npx.
https://www.npmjs.com/package/npx

I would like to get support for running this package at the plugin level.

rochakgupta:
Is anyone up for it?

iamruslanbakirov:
@rochakgupta 

i have simple workaround:

you need create new npm script, and executed with frontend-maven-plugin

example npm script:
`node node/node_modules/npm/bin/npx-cli.js --package YOU_DEPENDECY@X.Y.Z --userconfig .npmrc COMMAND`

`node node/node_modules/npm/bin/npx-cli.js --package webpack-cli@3.1.0 --userconfig .npmrc webpack`

kakarlus:
@iamruslanbakirov 
i tried your workaround, works like a charm, but i didn't have to provide the --package flag.
i just straight replaced `npx` directly with `node node/node_modules/npm/bin/npx-cli.js`

Ginxo:
@kakarlus @rochakgupta can you please add here the "execution" configuration for this?
I tried this
```
          <execution>
            <id>npm treat lock files</id>
            <phase>compile</phase>
            <goals>
              <goal>npm</goal>
            </goals>
            <configuration>
              <arguments>node node/node_modules/npm/bin/npx-cli.js --userconfig .npmrc npmtt remove</arguments>
            </configuration>
          </execution>
```
with this result
```
[INFO] Running 'npm node node/node_modules/npm/bin/npx-cli.js --userconfig .npmrc npmtt remove' in /home/user/project
[INFO] 
[INFO] Usage: npm <command>
[INFO] 
[INFO] where <command> is one of:
[INFO]     access, adduser, audit, bin, bugs, c, cache, ci, cit,
[INFO]     clean-install, clean-install-test, completion, config,
[INFO]     create, ddp, dedupe, deprecate, dist-tag, docs, doctor,
[INFO]     edit, explore, get, help, help-search, hook, i, init,
[INFO]     install, install-ci-test, install-test, it, link, list, ln,
[INFO]     login, logout, ls, org, outdated, owner, pack, ping, prefix,
[INFO]     profile, prune, publish, rb, rebuild, repo, restart, root,
[INFO]     run, run-script, s, se, search, set, shrinkwrap, star,
[INFO]     stars, start, stop, t, team, test, token, tst, un,
[...]

PR #867, opened 2019-12-17
Issue #750: Add support for npx runner

jimmyfm:
Trying to address: #750 

We internally use "license-checker", but since we are mostly a java shop we build our react ui with this mvn plugin. We wanted to have license checking integrated with the build configuration without having to rely on workarounds.

The change in this request allow our configuration to work. I know it may need some adjustments and testing, and I'd be happy to gather feedback and integrate it.

Little slice of our pom.xml:
```
<execution>
	<id>generate license list</id>
	<goals>
		<goal>npx</goal>
	</goals>
	<configuration>
		<arguments>license-checker --production --csv</arguments>
	</configuration>
</execution>
```

### DIFF
--- a/frontend-maven-plugin/src/it/npx-integration/package.json
+++ b/frontend-maven-plugin/src/it/npx-integration/package.json
@@ -1,0 +1,5 @@
+{
+  "name": "example",
+  "version": "0.0.1",
+  "dependencies": { }
+}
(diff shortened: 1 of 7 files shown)


eirslett:
Looks good, can you try adding some integration tests as well?
Originally I wanted to limit the number of "custom" runners, but I think npx might be a good exception to the rule.

jimmyfm:
I can't replicate locally the failures I am seeing in travisci. How can I fix that?

In the code I can find only two test classes, where are the integration tests I should add to?

eirslett:
> I can't replicate locally the failures I am seeing in travisci. How can I fix that?

I think it fails because of Java 11, maybe?

> In the code I can find only two test classes, where are the integration tests I should add to?

https://github.com/eirslett/frontend-maven-plugin/tree/master/frontend-maven-plugin/src/it

Issue #860, opened 2019-11-22
Use the provided npm from Node archive when npmVersion is not set


ndywicki:
Hi team,

We are behind a corporate firewall that block all `.exe` files and on our windows dev OS we can't install the NodeJs binary.

I see that if the `npmVersion` attribut is not set the plugin use instead of the win64 `*.tar.gz` and put the `node.exe` in `./node` directory but not the `npm` binaries (npm and npm.cmd).

See [NPMInstaller](https://github.com/eirslett/frontend-maven-plugin/blob/1fdf2102b975ff52bbe0ce7aece5eb4f809e31e0/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NPMInstaller.java#L76)

Consequently plugin use the global npm bin and there are somes warn :

````
[INFO] npm WARN lifecycle The node binary used for scripts is C:\workspaces\sample\node\node.exe 
but npm is using C:\Program Files\nodejs\node.exe itself. Use the `--scripts-prepend-node-path` option to include the path for the node binary npm was executed with.
````

I make a trick to use a maven Ant task copy and play with the Maven lifecycles phase:

````xml
<copy file="node/node_modules/npm/bin/npm" todir="node"/>
<copy file="node/node_modules/npm/bin/npm.cmd" todir="node"/> 
````

but that's not the ideal.
Can you add an option to use and copy the npm binaries provided by the node archive even if the `npmVersion` is not set?

Regards,
Nicolas

eirslett:
It's possible to upload the binaries to your internal Nexus repository (if you have one?) and download them from there instead, to circumvent the firewall.
Otherwise, I would suggest making an opening in the firewall, for fetching these binaries.

Issue #458, opened 2016-08-30
Use npm version provided with node if not specified


ufoscout:
Hi all,

The current configuration of the install-node-and-npm goal requires both the node and the npm versions. This forces to manually check what is the npm version provided with a specific node release and to update both the values. However, nodejs is already packaged with a specific version of npm. Would it be possible to use the npm version provided by default by node if the npmVersion is omitted?

For example, in this case:

```
    <execution>
        <id>install node and npm</id>
        <goals>
            <goal>install-node-and-npm</goal>
        </goals>
        <configuration>
            <nodeVersion>v6.2.2</nodeVersion>
            <npmVersion>3.9.5</npmVersion>
        </configuration>
    </execution>
```

it will use node 6.2.2 and npm 3.9.5. While in this case:

```
    <execution>
        <id>install node and npm</id>
        <goals>
            <goal>install-node-and-npm</goal>
        </goals>
        <configuration>
            <nodeVersion>v6.2.2</nodeVersion>
        </configuration>
    </execution>
```

it will use node 6.2.2 and whatever npm version provided with it.

eirslett:
In the Windows case, npm isn't packaged with node, it has to be downloaded separately.
(At least that was the case when the plugin was written)

stevenbower:
I have a patch that allows for this... Will post shortly.. It does look like NPM is provided from around the 4.x line and above on all platforms.. (i've confirmed on windows/linux but it looks from their docs this will be the case going forward).

ndywicki:
Hi Guys,

We use the npm provided mode because we are behind a coorporate firewall who block node.exe download.
So without specify the `npmVersion` properties the plugin download the node tar.gz correctly.
It's great but the [npm install code](https://github.com/stevenbower/frontend-maven-plugin/commit/49b77ecc330873ea710dd2885076b1fda5f6cff7#diff-4b3b75f0c89bc476e9c26111e6a1e832R242) not copy the binaries under the node root directory like node.exe.

Result we have some warrning because it's the npm system used instead of the local project version:

```
[INFO] npm WARN lifecycle The node binary used for scripts is C:\workspaces\sample\node\node.exe 
but npm is using C:\Program Files\nodejs\node.exe itself. 
Use the `--scripts-prepend-node-path` option to include the path for the node binary npm was executed with.
```

Thanks

It's ok if we manually copies the binaries from `node/node_module/npm/bin/[npm,npm.cmd]` into `node/`

PR #868, opened 2019-12-21
Copying npm scripts for embedded node

mwatts15:
**Summary**
If the "provided" npm version is used, the binaries are not copied into the node directory so they aren't in the path for `npm` executions started from the plugin. For example, the `prebuild` in the example project calls `npm install` which will use the system npm rather than that installed for the porject. This is experienced by reporters of #458 and #860.

**Tests and Documentation**
Modified existing node-provided-npm integration test.

The example test is failing due to some phantomjs error and is apparently unrelated.


eirslett:
Looks good to me, thanks! I guess it'll be in the christmas release! :-D

eirslett:

PR #876, opened 2020-01-23
Added option for specifying installers locally instead of downloading…

akang:
The corporate policy where I work makes it hard for me to download anything from any unauthorized urls into the build servers even by proxy. 

So by having these options,we can store the node and npm installers locally and skip the need to download anything from the Internet. 

The downside is that you will have to manually copy the node and npm installers into version control but certainly better that not being able to run this wonderful plugin.


eirslett:
This use case has been discussed some times - most people end up just using [exec-maven-plugin](https://www.mojohaus.org/exec-maven-plugin/) instead, which is really an easier option if you're going to commit the binaries into your repository anyways!

Issue #887, opened 2020-03-10
Npm Non Proxy Hosts doc and support


arykov:
As of NPM v6.4.1 (2018-08-22) statement "npm does not currently support non proxy hosts" is no longer accurate. Changelog is [here](https://github.com/npm/cli/blob/latest/CHANGELOG.md#v641-2018-08-22). Updated documentation is [here](https://docs.npmjs.com/misc/config#noproxy).

Any appetite for the change to support it? I can take a crack at it.

PR #891, opened 2020-03-27
Added No Proxy Support

danielle-mustillo:
**Summary**
NPM now supports setting noproxy hosts, which is a field that is often set
by users in their setting.xml proxy settings. We now push that all the way
down to the npm calls.

This should enable users to inherit their settings.xml proxy settings faithfully,
avoiding the need to use the `npmInheritsProxyConfigFromMaven` setting in
these cases at least.

This also fixes https://github.com/eirslett/frontend-maven-plugin/issues/887

**Tests and Documentation**
No tests were necessary as it is fairly trivial and non seem to exist presently. I can write some if desired.

### DIFF
--- a/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NpmRunner.java
+++ b/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NpmRunner.java
@@ -37,6 +37,7 @@ final class DefaultNpmRunner extends NodeTaskExecutor implements NpmRunner {
 
             arguments.add("--https-proxy=" + proxy.getUri().toString());
             arguments.add("--proxy=" + proxy.getUri().toString());
+            arguments.add("--noproxy=" + proxy.getNonProxyHosts());
         }
         
         return arguments;
--- a/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NpmRunner.java
+++ b/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NpmRunner.java
@@ -14,7 +14,8 @@ final class DefaultNpmRunner extends NodeTaskExecutor implements NpmRunner {
         super(config, TASK_NAME, config.getNpmPath().getAbsolutePath(), buildArguments(proxyConfig, npmRegistryURL));
     }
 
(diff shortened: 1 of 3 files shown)


frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NpmRunner.java
```diff
@@ -37,6 +37,7 @@ public DefaultNpmRunner(NodeExecutorConfig config, ProxyConfig proxyConfig, Stri
 
             arguments.add("--https-proxy=" + proxy.getUri().toString());
             arguments.add("--proxy=" + proxy.getUri().toString());
+            arguments.add("--noproxy=" + proxy.getNonProxyHosts());
```
ricksbrown, line 40: To handle cases where nonProxyHosts is empty or absent perhaps this needs a null/empty check to avoid `--noproxy=null` or `noproxy=  ` ?

danielle-mustillo: I added the check, and I added a unit test for that check (as well as the entire `buildArguments` method)

frontend-plugin-core/src/test/java/com/github/eirslett/maven/plugins/frontend/lib/DefaultNpmRunnerTest.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    private List<String> runBuildArguments(String nonProxyHost, String registryUrl) {
+        List<ProxyConfig.Proxy> proxyList = Stream.of(
+                new ProxyConfig.Proxy(id, protocol, host, port, username, password, nonProxyHost)
+        ).collect(Collectors.toList());
+
+        return DefaultNpmRunner.buildArguments(new ProxyConfig(proxyList), registryUrl);
+    }
+}
```
ricksbrown, line 83: nitpick - all the other files have a newline at the end...

eirslett: It doesn't matter too much, I guess...

eirslett:
Thanks! Sorry I didn't notice this PR earlier!

PR #903, opened 2020-06-03
Fix testFailureIgnore config in readme

tarioch:
**Summary**
Readme is wrong about the config option for enabling/disabling test failure ignore setting. The config is different from the property name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -470,7 +470,7 @@ tag of an execution like this:
 
 ```xml
 <configuration>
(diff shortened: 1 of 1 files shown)


eirslett:
Are you sure they're not mapped to the same property?

eirslett:
It's like this in the source code:
```
    @Parameter(property = "maven.test.failure.ignore", defaultValue = "false")
    protected boolean testFailureIgnore;
```

tarioch:
This is about the config not the property.

Issue #913, opened 2020-06-30
frontend not uses the latest gulp-cli


Zardoz89:
frontend-maven-plugin uses gulp-cli from "node_moules/gulp/node_modules" .

**Why this is a problem ?** 
Well, it not allow to use the latest version of gulp-cli that supports using ES6 modules on gulpfiles.

**What is the expected behavior?**

Should try first to use "node_moules/gulp-cli". If isn't found, then should use the gulp-cli included with gulp, like actually does.

PR #914, opened 2020-06-30
Resolves issue #913 . Try to use gulp-cli first

Zardoz89:
**Summary**

Resolves issue #913 . First try to use gulp-cli if it's found on node-modules. If it not exists, then fallbacks to the old behaviour where calls to "node-modules/gulp/bin/gulp.js" that calls the embed gulp-cli of gulp.js

**Tests and Documentation**

Given this two files : 

gulpfile.js
```
import gulp from "gulp";

const { src, dest } = gulp;

export default() => {
  src('input/*.js')
    .pipe(dest('output/'))
}
```

package.js
```
{
  "main": "gulpfile.js",
  "devDependencies": {
    "gulp": "^4.0.2",
    "gulp-cli": "^2.3.0"
  },
  "type": "module"
}
```

It must run correctly and copy any js file from input to output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Last public release: [![Maven Central](https://maven-badges.herokuapp.com/maven-
 
 ## Changelog
 
+### 1.10.1
+
+* Gulp goal tries to use gulp-cli if is installed as direct dependency (#913)
+
 ### 1.9.0
 
 * Copy npm scripts, so they are available for execution ([#868](https://github.com/eirslett/frontend-maven-plugin/pull/868))
--- a/README.md
+++ b/README.md
@@ -264,7 +264,7 @@ By default, colors will be shown in the log.
 ### Running gulp
 
 Very similar to the Grunt execution. It will run gulp according to the `gulpfile.js` in your working directory.
-By default, colors will be shown in the log.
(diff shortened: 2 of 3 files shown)


eirslett:
Actually, I don't think people should be using the gulp mojo - a much better alternative is to use the `npx` mojo/goal, to run gulp. That way, it will work with this Maven plugin in the future, even if gulp breaks in some other way. The same applies to grunt and webpack as well.

Zardoz89:
I think that it's a good idea, but should be done on another issue/pull request (as need more changes and proper testing).
This could be merged as is it now (and I need ASAP, so I could rewrite my gulpfiles to use ES6 import/export)

eirslett:
What I mean is, you can do it today with this version of frontend-maven-plugin (it's already supported), you just configure it differently!

Instead of this:
```
<execution>
    <id>gulp build</id>
    <goals>
        <goal>gulp</goal>
    </goals>
    <configuration>
        <arguments>build</arguments>
    </configuration>
</execution>
```
you can do this:
```
<execution>
    <id>gulp build</id>
    <goals>
        <goal>npx</goal>
    </goals>
    <configuration>
        <arguments>gulp build</arguments>
    </configuration>
</execution>
```

Zardoz89:
Oh! Sorry I didn't understand you the first time. I did a little test, and works fine for my case.

PR #919, opened 2020-07-08
Replace | with , for noproxyHosts

### LINKED ISSUE #918: No Proxy Settings from settings.xml in wrong format
Report a Bug.

in settings.xml are nonProxyHosts defined according to http://maven.apache.org/guides/mini/guide-proxies.html

npm is called with `--noproxy=localhost|127.0.0.1|192.*` 
The npm documentation https://docs.npmjs.com/misc/config#noproxy
states `A comma-separated string or an array of domain extensions that a proxy should not be used for.`
So both formats are not directly interchangeable.

Have nonProxyHosts set in your settings.xml with multiple hosts

a correct noproxy parameter for npm

Win 10 1904, frontend-maven-plugin:1.10.0

tkalmar:
This should fix the newly introduced noproxy handling ( fixes #918)

**Summary**

**Tests and Documentation**


eirslett:
Thanks!

Issue #880, opened 2020-02-05
npm with arguments runs forever


garlikus:
**Do you want to request a _feature_ or report a _bug_?**
Bug report.

**What is the current behavior?**
Not all arguments are passed to npm resulting in plugin execution running forever. It doesn't work for: `run ng run `. Apparently the second run is dismissed. I suspect `com.github.eirslett.maven.plugins.frontend.lib.ArgumentsParser#addArgument` to cause the problem.

**If the current behavior is a bug, please provide the steps to reproduce.**
`<execution>

	<id>web-app-tests</id>
	<goals>
		<goal>npm</goal>
	</goals>
	<configuration>
		<arguments>run ng run web-app:test:ci</arguments>
	</configuration>
</execution>`

**What is the expected behavior?**
All arguments defined in `arguments` are passed to npm.

**Please mention your frontend-maven-plugin and operating system version.**
Doesn't work since version 1.7.4 upwards.
The last version it worked is 1.6.

eirslett:
Try `<arguments>run ci-tests</arguments>` and in your `package.json` file, create an entry `ci-test` in the `scripts` field?

pankajguptain:
Hi, I need to pass an environment variable to run command like ng build --configuration=uat. How can I achieve the same.  If I am running below script, it is just doing ng build and picking up development configuration but I want to pick uat configuration defined in angular.json. 

   <execution>
         <id>npm run build</id>
   <goals>
            <goal>npm</goal>
   </goals>
   <configuration>
           <arguments>run build</arguments>
    </configuration>
   </execution>

Thanks in advance

josephw:
@garlikus, does #930 address this problem?

PR #930, opened 2020-09-08
Accept repeated values for regular arguments.

josephw:
The arguments you want to pass may repeat a value. Accept this case, but continue to ignore additionalArguments that have already been supplied.

**Summary**

Accept repeated values for arguments. I have a use case of:

```
<arguments>--arg1 Value --arg2 Value</arguments>
```

and, currently, this is passed through as `--arg1 Value --arg2`.

As a workaround, I could use another form (e.g., `--arg1=Value --arg2=Value`), but I'm aiming to upgrade from a pre-`ArgumentsParser` version of this plugin, and it would be good to leave usage as-is.

(I believe this is also the bug reported in #880.)

**Tests and Documentation**

Existing unit tests pass. A new test demonstrates the new behaviour. I don't believe the original behaviour was documented, so this doesn't make any documentation changes.

### DIFF
--- a/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ArgumentsParser.java
+++ b/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ArgumentsParser.java
@@ -65,7 +65,9 @@ class ArgumentsParser {
         addArgument(argumentBuilder, arguments);
 
         for (String argument : this.additionalArguments) {
-            addArgument(argument, arguments);
+            if (!arguments.contains(argument)) {
+                arguments.add(argument);
+            }
         }
 
         return new ArrayList<>(arguments);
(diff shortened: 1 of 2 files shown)


josephw:
Looking through open PRs; this is an identical fix to #810, which was opened back in May 2019, but didn't get a response. The code change being so similar makes me more confident that this is the right fix. That PR has conflicts due to also updating `CHANGELOG.md`, but would otherwise be good to merge.

@eirslett, any feedback on whether this is a reasonable approach? My aim is to move off an old internal fork of this plugin.

josephw:
Hey @eirslett; does this (or #810) look good to merge? I'm happy to make revisions, or consider another approach.

josephw:
@eirslett ping; some response welcome 😄. My alternative is to update and maintain a fork, which doesn't seem like valuable work.

eirslett:
Hey, sorry! I'll merge it now!

josephw:
Thanks!

PR #934, opened 2020-10-03
Supports Alpine by detecting OS using /etc/alpine-release

adriancole:
**Summary**

Alpine is a small OS distribution that supports JDK. Supporting Alpine
means less download time and cost when building projects in Docker.

This solves the same problem #853 did, via a different approach.

* Move the default download root to the `Platform` impl
* Allow a classifier (currently only musl, but there are others)
* Special case Alpine instead of generically

The above changes allow someone to override the download root regardless
of if the nodejs dist is experimental or not.

**Tests and Documentation**

This backfills tests for `Platform` using the same approach as we do in Brave
(powermock).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Last public release: [![Maven Central](https://maven-badges.herokuapp.com/maven-
 
 ## Changelog
 
+### 1.10.0
+
+* Supports Alpine Linux
+
 ### 1.9.0
 
 * Copy npm scripts, so they are available for execution ([#868](https://github.com/eirslett/frontend-maven-plugin/pull/868))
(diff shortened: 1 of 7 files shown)


adriancole:
testing manually here (our builder image is alpine) https://github.com/openzipkin/zipkin/compare/DO_NOT_MERGE_test-frontend-maven-plugin?expand=1

adriancole:
^^ works fine!

eirslett:
OK, let's go with this implementation!

eirslett:
There we go, it's in the 1.10.2 release.

adriancole:
Thanks for the quick turnaround, @eirslett!

PR #935, opened 2020-10-07
Fix proxy parsing

tkalmar:
**Summary**

After fixing the proxy parsing (replacing | with ,) a friend of mine pointed me towards: https://npm.community/t/noproxy-setting-with-more-than-one-hostname/5054/3
So the comma separated String seems not to work from commandline. Hence this fix adds an extra parameter for every proxy exclusion.
I also noted that the npm documentation not mentions wildcards (seems to be a java thingy), so they are also removed when parsing the proxy settings.

**Tests and Documentation**


mickroll:
Please integrate as soon as possible, we really depend on working proxy configuration.

eirslett:
Ok, I guess it looks good!

PR #949, opened 2020-12-15
Download x64 binary on Apple Silicon

ajthom90:
**Summary**

I have been testing out an Apple Silicon Mac mini. I am using the Azul Zulu arm64 OpenJDK (JDK 8). This change tells the plugin to download the x64 version of Node on macOS, which should run fine under Rosetta. 

I'm happy to write up an issue also, if you so desire! Thank you!

**Tests and Documentation**

When building using Maven on Apple Silicon, I receive an error, as the URL generated by the plugin doesn't work, since there is not currently (as of 12/15/2020) an arm64 version of Node for macOS. So, this uses a simple check when generating the download URL: if the OS is Mac, and the architecture is arm64, download the x64 binary.

```
[INFO] 
[INFO] --- frontend-maven-plugin:1.9.1:install-node-and-npm (install-npm) @ <<PROJECT NAME>> ---
[INFO] Installing node version v11.9.0
[INFO] Downloading https://nodejs.org/dist/v11.9.0/node-v11.9.0-darwin-arm64.tar.gz to /Users/localuser/.m2/repository/com/github/eirslett/node/11.9.0/node-11.9.0-darwin-arm64.tar.gz
[INFO] No proxies configured
[INFO] No proxy was configured, downloading directly
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 1.065 s
[INFO] Finished at: 2020-12-15T13:10:54-08:00
[INFO] Final Memory: 29M/359M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal com.github.eirslett:frontend-maven-plugin:1.9.1:install-node-and-npm (install-npm) on project <<PROJECT NAME>>: Could not download Node.js: Got error code 404 from the server. -> [Help 1]
```

### DIFF
--- a/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/Platform.java
+++ b/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/Platform.java
@@ -135,7 +135,12 @@ class Platform {
     }
 
     public String getNodeClassifier() {
-        String result = getCodename() + "-" + architecture.name();
+        final String result;
+        if(isMac() && architecture == Architecture.arm64) { // this check is required to download the x64 binary until there is an arm64 version available for macOS (darwin).
+            result = getCodename() + "-" + Architecture.x64.name();
+        } else {
+            result = getCodename() + "-" + architecture.name();
+        }
         return classifier != null ? result + "-" + classifier : result;
     }
 }
--- a/frontend-plugin-core/src/test/java/com/github/eirslett/maven/plugins/frontend/lib/PlatformTest.java
+++ b/frontend-plugin-core/src/test/java/com/github/eirslett/maven/plugins/frontend/lib/PlatformTest.java
@@ -32,6 +32,18 @@ public class PlatformTest {
     }
 
     @Test
+    public void detect_arm_mac_download_x64_binary() {
+        mockStatic(OS.class);
+        mockStatic(Architecture.class);
+
+        when(OS.guess()).thenReturn(OS.Mac);
+        when(Architecture.guess()).thenReturn(Architecture.arm64);
+
+        Platform platform = Platform.guess();
+        assertEquals("darwin-x64", platform.getNodeClassifier());
+    }
+
+    @Test
     public void detect_linux_notAlpine() throws Exception {
         mockStatic(OS.class);
         mockStatic(Architecture.class);


eirslett:
Ok, thanks!

eirslett:
Should be released now, in version 1.11.0. Could you follow up this whenever Node.js publishes arm64 builds? Then we could switch back, if that's worth the extra work.

ajthom90:
Absolutely. I'll put a reminder on my calendar to check weekly. Thanks for your quick reply on this, I appreciate it!

grishka:
Uhh sorry but why is this a thing? As of right now, there are native ARM macOS builds available using that exact URL format:

https://nodejs.org/dist/v16.13.1/node-v16.13.1-darwin-arm64.tar.gz

Could you please undo this PR?

eirslett:
> Uhh sorry but why is this a thing? As of right now, there are native ARM macOS builds available using that exact URL format

It's because @ajthom90 should check weekly if there was a native arm64 build :wink:

Since when did they start publishing Silicon builds? Maybe we need some version smartness in the plugin, so that it downloads the x64 one for older versions of Node, and arm64 for newer versions of Node.

grishka:
> Since when did they start publishing Silicon builds?

No idea. I just got my M1 macbook last Friday and noticed that this plugin downloads an x86 build, while the standalone node I installed is universal.

> Maybe we need some version smartness in the plugin, so that it downloads the x64 one for older versions of Node, and arm64 for newer versions of Node.

Yeah, that makes sense in case someone needs an older version. Maybe I should even send a PR of my own 🤔

grishka:
It appears that they have started providing arm64 builds since 16.0.0. That's convenient for comparison. I'll submit a PR.

grishka:
image
Huh? It's already done in this exact way in commit 4a63273262e9bc69b9123e0e67885c33e59cc7a3 🤔🤔🤔🤔🤔

Sorry, I'm stupid. I just needed to update the node version in my pom.xml to something newer.

ajthom90:
> Huh? It's already done in this exact way in commit [4a63273](https://github.com/eirslett/frontend-maven-plugin/commit/4a63273262e9bc69b9123e0e67885c33e59cc7a3) 🤔🤔🤔🤔🤔

> It's because @ajthom90 should check weekly if there was a native arm64 build 😉

Haha whoops! Glad that @grishka spotted that this was already fixed! I'd stopped checking for arm64 native node builds because I remembered that commit! Thanks all!

Issue #996, opened 2021-08-30
Need a new Mojo to execute arbitrary node command


cmarchand:
Some frameworks need to pre-compile some files before making the front app available. See [xslt3](https://www.npmjs.com/package/xslt3) for example where we need to pre-compile xslt files with this commande :
```
node xslt3 -xsl:whatever.xsl -export:whatever.sef.json
```

Actually, there is no Mojo that directly allows to call `node` with specific parameters. I need one, with this usage :
```xml
<plugin>
  <artifactId>frontend-maven-plugin</artifactId>
  <executions>
    <execution>
      <id>compile-xslt</id>
      <phase>compile</phase>
      <goals>
        <goal>node</goal>
      </goals>
      <configuration>
       <arguments>node_modules/xslt3/xslt3.js -xsl:src/main/xsl/md-to-xhtml.xsl -export:target/classes/xsl/md-to-html.sef.json</arguments>
      </configuration>
    </execution>
  </executions>
</plugin>
```

I just open the issue to be referenced in the pull request.

zacthompson:
There is the `npx` goal which would probably work almost exactly as you want: `<arguments>xslt3 -xsl:...`

You might also want to consider https://docs.npmjs.com/cli/v6/using-npm/scripts -- in particular I'm guessing that you want to do this xslt call as part of the preinstall or postinstall of your package or something like that. But even if not, you can add it as its own script and then make the f-e-p goal `npm` and the arguments `run compile-xslt`.

chulkilee:
FYI `npm run` of npm 7.x has a strange behaviour

https://docs.npmjs.com/cli/v7/using-npm/scripts#user

> When npm is run as root, scripts are always run with the effective uid and gid of the working directory owner.

As a result, it runs as different user from maven... hitting the permission issue 🤦 

Right now I'm just using a script to set up PATH with node/npm installed from this plugin. 

See https://github.com/npm/cli/issues/3110

PR #997, opened 2021-08-30
New Mojo to run node with arbitrary arguments

cmarchand:
**New Mojo to execute arbitrary node commands**

As explained in issue #996 I need a new Mojo to run `node` command with arbitrary parameters

**Tests and Documentation**

README.md contains an example to use NodeMojo plugin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -408,9 +408,11 @@ You may need to run `node` with arbitrary arguments. Check this Mojo :
     <goals>
         <goal>node</goal>
     </goals>
-
+    
+    <!-- optional: the default phase is "compile" -->
     <phase>compile</phase>
 
+    <!-- required: arguments to launch node with -->
     <configuration>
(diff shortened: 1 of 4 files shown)


eirslett:
A few people have asked for this feature, but perhaps not many enough. What people usually do is `npm run your-script` from Maven, and then in package.json add `"your-script": "node something.js"` inside the `scripts` block. You then get the additional benefit of being able to do `npm run your-script` directly from the command line (without Maven) which can be convenient if your command is somewhat complex (`node some/kind/of/script.js --foo bar ./bla --etc`).

cmarchand:
Yes, that's a solution. In my situation, at this point, I do not have anymore need of npm, so adding a `package.json` in my maven project is not very useful.
But yes, it's a good workaround. Thanks.

Issue #1002, opened 2021-10-01
npx not in PATH when running as part of npm goal


eidottermihi:
**Do you want to request a _feature_ or report a _bug_?**

Bug

**What is the current behavior?**

Trying to build a project with a package.json that uses npm-force-resolutions via npx in `preinstall`:

```json
{
  "name": "blabla",
  "scripts": {
    "preinstall": "npm install --package-lock-only --ignore-scripts && npx npm-force-resolutions",
   }
}
```

using the following Maven configuration:

```xml
<plugin>
	<groupId>com.github.eirslett</groupId>
	<artifactId>frontend-maven-plugin</artifactId>
	<executions>
		<execution>
			<goals>
				<goal>install-node-and-npm</goal>
			</goals>
		</execution>
                <execution>
		    <id>npm-install</id>
		    <phase>generate-sources</phase>
		    <goals>
			    <goal>npm</goal>
		    </goals>
		    <configuration>
			    <arguments>install</arguments>
		    </configuration>
	    </execution>
	</executions>
</plugin>
```

fails when running in a GitLab CI build using official docker image `maven:3-jdk-11`

```
[INFO] --- frontend-maven-plugin:1.12.0:install-node-and-npm (install node and npm) @ myproject ---
[INFO] Installing node version v14.16.0
[ ..... ]
[INFO] Extracting NPM
[INFO] Installed node locally.
[INFO] 
[INFO] --- frontend-maven-plugin:1.12.0:npm (npm-install) @ myproject  ---
[INFO] Running 'npm install' in /builds/fsQupcmA/0/group/project/repo/module
[INFO] 
[INFO] > mypackage@0.1.0 preinstall /builds/fsQupcmA/0/group/project/repo/module
[INFO] > npx npm-force-resolutions
[INFO] 
[INFO] sh: 1: npx: not found
[INFO] npm ERR! code ELIFECYCLE
[INFO] npm ERR! syscall spawn
[INFO] npm ERR! file sh
[INFO] npm ERR! errno ENOENT
```

**If the current behavior is a bug, please provide the steps to reproduce.**

Use a package.json that has a preinstall script that uses npx.

**What is the expected behavior?**

IMO npx should be on the PATH when `npm` goal is executed.

**Please mention your frontend-maven-plugin and operating system version.**

frontend-maven-plugin: 1.12.0
OS: Linux x64 / Docker Image maven:3-jdk-11 (https://hub.docker.com/_/maven)

PR #1007, opened 2021-10-31
Copying npx scripts for embedded node

vzhyhunou:
**Summary**

Copies npx scripts along with npm ones to be able to run preinstall script from package.json that uses npx.

Resolves [#1002](https://github.com/eirslett/frontend-maven-plugin/issues/1002)


eirslett:
Thanks! 🥳

rmueller83:
Would it make sense to also add this to `NodeInstaller.java`? We currently suffer from a failing build that uses yarn and a plugin that calls `npx` during postinstall:
https://github.com/TroyAlford/react-jsx-parser/blob/develop/package.json#L83

PR #1012, opened 2021-11-20
Added Yarn Berry compatibility flag

### LINKED ISSUE #928: Support for Yarn 2
Does `frontend-maven-plugin` support Yarn 2? [Yarn's migration guide for Yarn 2](https://yarnpkg.com/advanced/migration) is not really clear to me, but mentions Yarn 2 is 'a very different software from the v1'.

If `frontend-maven-plugin` does not support Yarn 2 yet, this is a feature request.

Yarn v1 works fine, but I don't know how to use Yarn v2.

frontend-maven-plugin 1.10.0 on macOS 10.15.6.

neo1203:
**Summary**

Fixes #928.
Added a compatibility flag for Yarn 2.x and above (Berry) so maven-frontend-plugin doesn't consider reported version as invalid anymore and doesn't try to re-install Yarn every time it runs.

**Tests and Documentation**

Updated CHANGELOG.md and README.md with details about the new flag

### DIFF
--- a/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/InstallNodeAndYarnMojo.java
+++ b/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/InstallNodeAndYarnMojo.java
@@ -1,5 +1,7 @@
 package com.github.eirslett.maven.plugins.frontend.mojo;
 
+import java.io.File;
+import java.util.stream.Stream;
 import org.apache.maven.execution.MavenSession;
 import org.apache.maven.plugins.annotations.Component;
 import org.apache.maven.plugins.annotations.LifecyclePhase;
@@ -15,6 +17,8 @@ import com.github.eirslett.maven.plugins.frontend.lib.YarnInstaller;
 
 @Mojo(name = "install-node-and-yarn", defaultPhase = LifecyclePhase.GENERATE_RESOURCES, threadSafe = true)
 public final class InstallNodeAndYarnMojo extends AbstractFrontendMojo {
+
+    private static final String YARNRC_YAML_FILE_NAME = ".yarnrc.yml";
 
     /**
      * Where to download Node.js binary from. Defaults to https://nodejs.org/dist/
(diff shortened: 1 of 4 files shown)


neo1203:
@eirslett , any feedback on this PR ?

eirslett:
Can we somehow auto-detect `isYarnBerry` instead of specifying it in the pom?

neo1203:
That's a nice suggestion.
We can detect the .yarnrc.yml file in the project folder.
In that case, I would still suggest to keep that additional `isYarnBerry` flag to handle corner cases.
Possible values and behavior :
- Not present / empty tag : auto-detect .yarnrc.yml file in the project folder
- A boolean (true or false) : force enable / disable

Would that be OK with you ? If it is, I will update the PR with that implementation

eirslett:
I don't even think we need the force enable/disable. We could add it later on, if anybody really needs it (which I highly doubt). My concern is to keep the number of configuration parameters in the plugin as low as possible.

frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/InstallNodeAndYarnMojo.java
```diff
@@ -16,6 +18,8 @@
 @Mojo(name = "install-node-and-yarn", defaultPhase = LifecyclePhase.GENERATE_RESOURCES, threadSafe = true)
 public final class InstallNodeAndYarnMojo extends AbstractFrontendMojo {
 
+    private static final String YARNRC_YAML_FILE_NAME = ".yarnrc.yml";
```
timja, line 21: the README you changed says `yaml`, probably safer to check both file extensions

neo1203: Good catch, my mistake. According to the Yarn documentation (https://yarnpkg.com/configuration/yarnrc) the file has to be named `.yarnrc.yml`, not `.yanrc.yaml`
I'll update the README

eirslett:
thanks!

Issue #604, opened 2017-05-15
Use the node and npm local if version match


cesarlarsson:
If there a way to use the local npm and node version and no download and create the node folder with those files if we have already installed the version on my local environment?

eirslett:
Closing this issue as a duplicate of #54 #108 #177 #243 #273 #277 #340 #344 #440 #471 :wink:

sorin7486:
Your philosophy sucks, mate. I must of seen 20 issues opened about this and you keep saying no, just because if you do things a certain way everybody else is expected to. Yeah, I'll be using exec instead.

jollytoad:
That's exactly the point, if you already have the tools installed you don't need this, just do it with exec. The primary purpose of this plugin (as I understand it) is to download local isolated copies of the JS tools for projects whose primary build pipeline is based around Java tooling.

qacwnfq:
Why would you ever want to use local installation over the one the plugin provides? It's much more stable to control the versions in the pom.xml then to leave it up to devs to keep track.
I worked on a project with legagy code and new code where two modules required different node versions to run correctly for some time. Keeping the node versions in the pom.xml files really prevented a lot of headache. It allows you to update your huge projects module by module instead of doing it all at once and breaking everything.

ppkarwasz:
We use Node.js only to generate our website, which 99% of the times is built and deployed on a GitHub runner that comes with Node.js preinstalled.

While technically we could use `exec-maven-plugin` or a profile that switches between `exec-maven-plugin` and `frontend-maven-plugin`, we prefer to use the latter, since it guarantees that:

- Contributors can also generate the website, without necessarily having to install Node.js first.
- We have a reproducible website, since the version of Node.js is fixed in the POM. If the GitHub runner changes Node.js version, we'll still use the specified version.

However the usage of `frontend-maven-plugin` is not ideal, since we have around 50 MiB of additional data in our Maven cache.

@eirslett, would you be open to making `NodeExecutor` and other services in your plugin injectable Sisu components?

If `NodeExecutor` was managed by Maven's dependency injection, users could provide their own implementations that would look for `node` with the appropriate version on the system `PATH` first.

PR #1015, opened 2021-12-09
Add support to use the system wide installation of node.js

fabianfrz:
**Summary**

This allows to use the node.js version provided by the operating system (installed via pacman, apt, dnf, whatever).

* In a restricted environment, It is very uncommon that you can download arbitrary executables from all over the internet
* It saves bandwidth of the build server since we don't have to download the npm and node in every build

**Tests and Documentation**

I guess this is hard to test as it heavily depends on the underlying operating system and it makes no sense to mock everything away. However I updated all relevant documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Last public release: [![Maven Central](https://maven-badges.herokuapp.com/maven-
 
 * update Dependency: Jackson (2.13.0), Mockito (4.1.0), JUnit (5.8.1), Hamcrest (2.2; now a direct dependency)
 * remove Dependency: Powermock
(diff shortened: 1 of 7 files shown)


eirslett:
In your particular use case, the best fit is probably [exec-maven-plugin](https://www.mojohaus.org/exec-maven-plugin/), which will use the system node/npm.

I'm sorry to break the bad news, but this has been discussed a number of times before, and will not be implemented. The reasons are explained in the other GitHub issues.

https://github.com/eirslett/frontend-maven-plugin/issues/604

> In a restricted environment, It is very uncommon that you can download arbitrary executables from all over the internet

The plugin supports proxies, so that you can host the executables locally behind your firewall (for example in a local Nexus setup)

> It saves bandwidth of the build server since we don't have to download the npm and node in every build

This is already fixed by the plugin - it will always cache the downloaded node/npm, they will never be fetched from the network again.

PR #1031, opened 2022-04-13
Add warning if determination of node version fails

kelunik:
**Summary**

We're currently facing an issue on macOS, which always installs node again, because it can't execute it correctly. Manual invocation of `<path>/node --version` works as expected. This will hopefully help to debug this.


eirslett:
Thanks!

Issue #1039, opened 2022-07-05
install-node-and-pnpm does not detect previously installed pnpm


nils-christian:
**Do you want to request a _feature_ or report a _bug_?**

This is a bug similar  to #519. 

**What is the current behavior?**

When running install-node-and-pnpm multiple times, it does not detect that pnpm was installed when run previously and instead installs pnpm from scratch each time. See log messages:

```
[INFO] --- frontend-maven-plugin:1.12.0:install-node-and-pnpm (default-cli) @ test ---
[INFO] Node v14.15.5 is already installed.
[INFO] PNPM 5.18.10 was installed, but we need version v5.18.10
[INFO] Installing pnpm version v5.18.10
```

**What is the expected behavior?**

pnpm should only be installed when it is either not installed or the version is actually different.

**Please mention your frontend-maven-plugin and operating system version.**

The frontend-maven-plugin-Version is 1.12.1, the OS is a windows based system.

nils-christian:
I created a PR to solve this issue: #1040

antonkai2k:
Seems like this ticket can be closed

PR #1040, opened 2022-07-06
[#1039] install-node-and-pnpm does not detect previously installed pnpm

nils-christian:
Fix for #1039.


eirslett:
I guess that should do the trick, thanks!

PR #1045, opened 2022-08-26
Fix #967 : Add pnpm executable to PATH

### LINKED ISSUE #967: Cannot use pnpm inside package.json scripts
Have a _package.json_ like:
```
..
"scripts": {
  "build": "pnpm build-a && pnpm build-b",
  "build-a": "webpack --config webpack.config.a.js",
  "build-b": "webpack --config webpack.config.b.js"
}
..
```
It is not possible to run _build_ with the corresponding pnpm maven goal, because pnpm executables are not added to PATH. Therefore child processes cannot be spawn with the embedded pnpm usage.

Although the PNPMInstaller tries to copy pnpm & pnpm.cmd ([copyPnpmScripts](https://github.com/eirslett/frontend-maven-plugin/blob/e727f7dbb1dfa1df0367aba43e93f88e5a8df342/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/PNPMInstaller.java#L170)), but pnpm has no such predefined files, thus it does not find anything to copy.

A solution can be perhaps to execute [cmd-shim](https://github.com/pnpm/cmd-shim) if the files are not found.
Or another approach can be to install the _.tgz_ files with npm instead of just extracting them, which I assume would add the executables automatically. But of course that would require the npm to be installed in advance.

JulesAaelio:
**Summary**

Fix the issue described in #967 

> the PNPMInstaller tries to copy pnpm & pnpm.cmd ([copyPnpmScripts](https://github.com/eirslett/frontend-maven-plugin/blob/e727f7dbb1dfa1df0367aba43e93f88e5a8df342/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/PNPMInstaller.java#L170)), but pnpm has no such predefined files, thus it does not find anything to copy.

The solution I propose is to create a symlink to `node_modules/pnpm/bin/pnpm.cjs` when no pnpm or pnpm.cmd is available. 

**Tests and Documentation**

I tested using this dummy project : https://github.com/JulesAaelio/test-frontend-maven-plugin

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,7 +26,7 @@ jobs:
           java-version: 1.8
       - name: "Run Maven"
         shell: bash
-        run: mvn clean install --batch-mode -PintegrationTests
+        run: mvn clean install --batch-mode -PintegrationTests -DintegrationTestProfiles=it-${{ runner.os }}
       - name: "Deploy"
         if: github.repository_owner == 'eirslett' && github.ref == 'refs/heads/master' && matrix.os == 'ubuntu-latest'
         shell: bash
--- a/frontend-maven-plugin/pom.xml
+++ b/frontend-maven-plugin/pom.xml
@@ -15,6 +15,10 @@
     <maven>3.6.0</maven>
   </prerequisites>
 
+  <properties>
+    <integrationTestProfiles>it-Linux</integrationTestProfiles>
+  </properties>
+
   <description>
         This Maven plugin lets you install Node/NPM locally
         for your project, install dependencies with NPM,
@@ -128,6 +132,7 @@
 
             <configuration>
               <debug>true</debug>
+              <profiles>${integrationTestProfiles}</profiles>
               <projectsDirectory>src/it</projectsDirectory>
               <cloneProjectsTo>${project.build.directory}/it</cloneProjectsTo>
               <settingsFile>src/it/settings.xml</settingsFile>
--- a/frontend-maven-plugin/src/it/pnpm-integration/pom.xml
+++ b/frontend-maven-plugin/src/it/pnpm-integration/pom.xml
@@ -41,19 +41,65 @@
                             <arguments>install</arguments>
                         </configuration>
                     </execution>
-
-                    <execution>
-                        <id>pnpm test</id>
-                        <goals>
-                            <goal>pnpm</goal>
-                        </goals>
-                        <configuration>
-                            <arguments>test</arguments>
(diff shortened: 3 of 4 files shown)


eirslett:
How will this work on Windows (which doesn't have symlinks, I think)?

GuillaumeC:
pnpm work with symlinks to a global store. So it should be work on Windows too.

JulesAaelio:
Hi @eirslett , 
You have a valid point here, although symlinks are supposed to work on Windows 10  it is still unclear under what conditions. 

According to their documentation, pnpm does'nt use symlinks neither, but junctions, that we cannot use here because we're trying to create a link to a file and not to a directory. https://pnpm.io/faq#does-it-work-on-windows

The others options to make this works on windows are :
As stated in the issue :  
> to execute [cmd-shim](https://github.com/pnpm/cmd-shim) if the files are not found.
> to install the .tgz files with npm instead of just extracting them, which I assume would add the executables automatically. But of course that would require the npm to be installed in advance.

Or : 
to use the freshly installed pnpm to "install" itself so it creates the executables. 

=> As all of these are non-trivial,  I propose to simply add a check before creating the symlink to ensure that we're running on a non-windows platform. I updated my code accordingly .

eirslett:
:+1:

However, this project has an integration test that tries to run pnpm. As far as I can see, that integration test is now passing. Is it possible to create a reproduction of this bug/issue (a failing integration test), to show that the problem has indeed been fixed with this patch?

JulesAaelio:
Sure, I can have a look. Should I include the new test in this merge request or in a new one ?

eirslett:
@JulesAaelio I merged your PR with the failing integration test. As you can see here: https://github.com/eirslett/frontend-maven-plugin/actions/runs/3169669688/jobs/5161739948 it looks like the integration test is still failing on Ubuntu? Can you have a look?

If you rebase your pull request branch, the integration test should be included in there, so it's easier to check whether the test passes or fails.

eirslett:
:+1: So how do we make the test pass on Windows then?

JulesAaelio:
As I said on my previous comment, making it work on windows is a bit more complicated and I don't feel like digging into it :) 
So I used a profile to ensure the test doesn't run. 
It is not really a fix but the tests are passing. Let me know what  you think :)

eirslett:
Ok, let's merge it anyways! And hopefully somebody on Windows will fix it if they need it.

PR #1053, opened 2022-10-01
Edit integration test to reproduce #967

JulesAaelio:
**Summary**

Edit integration test to reproduce #967 as requested in https://github.com/eirslett/frontend-maven-plugin/pull/1045


eirslett:
Great! I updated the GitHub Workflow so that it actually runs all integration tests on Windows, MacOS and Linux. Looks like the integration test is indeed failing!

Issue #657, opened 2017-10-11
cypress proxy problem


cforce:
[INFO] --- frontend-maven-plugin:1.6:install-node-and-yarn (install node and yarn) @ neo4j-browser ---
[INFO] Found proxies: [myproxy{protocol='http', host='10.175.249.97', port=8080}]
[INFO] Node v6.11.4 is already installed.
[INFO] Yarn 1.2.0 is already installed.
[INFO] 
[INFO] --- frontend-maven-plugin:1.6:npm (npm install) @ neo4j-browser ---
[INFO] Found proxies: [tsi{protocol='http', host='10.175.249.97', port=8080}]
[INFO] Running 'npm install --https-proxy=http://10.175.249.97:8080 --proxy=http://10.175.249.97:8080' in C:\Users\fwallwit\git\neo4j-browser
[INFO] 
[INFO] > cypress@0.20.3 postinstall C:\Users\fwallwit\git\neo4j-browser\node_modules\cypress
[INFO] > node index.js --exec install
[INFO] 
[INFO] Installing Cypress (version: 0.20.3)
[INFO] 
[INFO] [?25l[08:06:50]  Downloading Cypress     [started]
[INFO] [08:06:50]  Downloading Cypress     [failed]
[INFO] [08:06:50] â†’ The Cypress App could not be downloaded.
[INFO] 
[INFO] Please check network connectivity and try again:
[INFO] ----------
[INFO] 
[INFO] URL: https://download.cypress.io/desktop/0.20.3?os=win
[INFO] Error: getaddrinfo EAI_AGAIN download.cypress.io:443
[INFO] ----------
[INFO] 
[INFO] Platform: win32 (6.1.7601)
[INFO] Cypress Version: 0.20.3
[INFO] [?25hThe Cypress App could not be downloaded.
[INFO] 
[INFO] Please check network connectivity and try again:
[INFO] ----------
[INFO] 
[INFO] URL: https://download.cypress.io/desktop/0.20.3?os=win
[INFO] Error: getaddrinfo EAI_AGAIN download.cypress.io:443
[INFO] ----------
[INFO] 
[INFO] Platform: win32 (6.1.7601)
[INFO] Cypress Version: 0.20.3
[WARNING] npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@^1.1.1 (node_modules\sane\node_modules\fsevents):
[...]

PR #683, opened 2017-12-28
Add proxy environment variables for npm postinstalls

Cavva79:
**Summary**

Found that for postinstall script, such openlayers 4 has, in a proxy network, this plugin is not passing right the proxy to node.js running the script.

**Tests and Documentation**

This fixes issue #657.
I tested and it seems to work with openlayers 4.6.4

PR #1095, opened 2023-06-28
Fix#657 add proxy environment variables for npm postinstalls

josehenriqueventura:
From: https://github.com/eirslett/frontend-maven-plugin/pull/683

Summary

Found that for postinstall script, such openlayers 4 has, in a proxy network, this plugin is not passing right the proxy to node.js running the script.

Tests and Documentation

This fixes issue https://github.com/eirslett/frontend-maven-plugin/issues/657.
I tested and it seems to work with openlayers 4.6.4

Thanks @Cavva79


eirslett:
Thanks!

Issue #882, opened 2020-02-14
flaky installation on ci server


geyuqiu:
```
$ ./mvnw com.github.eirslett:frontend-maven-plugin:install-node-and-npm -DnodeVersion=v12.13.1 -DnpmVersion=6.13.4 -Dmaven.repo.local=$MAVEN_USER_HOME
[ERROR] The archive file .../.maven/com/github/eirslett/node/12.13.1/node-12.13.1-linux-x64.tar.gz is corrupted and will be deleted. Please try the build again.
[ERROR] Failed to execute goal com.github.eirslett:frontend-maven-plugin:1.8.0:install-node-and-npm (default-cli) on project cbkop-gui-unified: Could not extract the Node archive: Could not extract archive: '.../.maven/com/github/eirslett/node/12.13.1/node-12.13.1-linux-x64.tar.gz': EOFException -> [Help 1]
```

**Do you want to request a _feature_ or report a _bug_?**
bug
**What is the current behavior?**
because of a "failure" ci-setup, every push is built under the same directory
flaky behavior: most of the time successfully
**suggested fix**
the archive maybe corrupted before
and reinstallation not should be clearing the old outdated / corrupted files?
question: try `rm -rf .maven` before installation ?
**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**
runs smoothly every time
**Please mention your frontend-maven-plugin and operating system version.**
Red hat linux: 7
frontend-maven-plugin: 1.8.0

SystemOutPrint:
+!

simontunnat:
I have the same issue in our GitLab CI environment.

It only happens from time to time:
`
09:11:11:813 [ERROR] Failed to execute goal com.github.eirslett:frontend-maven-plugin:1.12.0:install-node-and-npm (install node and npm) on project one-ebmpapst.ui.frontend: Could not extract the Node archive: Could not extract archive: '/builds/it.ics.epm/one-intranet/one-ebmpapst/.m2/repository/com/github/eirslett/node/16.14.2/node-16.14.2-linux-x64.tar.gz': EOFException -> [Help 1]
[...]

PR #1098, opened 2023-07-04
Add configurable number of retries and interval between retries

snuyanzin:
The PR adds retry option of download and install for node, npm, yarn.

We (Flink community) sometimes faced an issue like 
```
[ERROR] The archive file /__w/1/.m2/repository/com/github/eirslett/node/16.13.2/node-16.13.2-linux-x64.tar.gz is corrupted and will be deleted. Please try the build again.
```
so archive was corrupted and removed. However it would be nice to have an autoretry for this case
i guess it could also partially address #882

### DIFF
--- a/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/InstallNodeAndNpmMojo.java
+++ b/frontend-maven-plugin/src/main/java/com/github/eirslett/maven/plugins/frontend/mojo/InstallNodeAndNpmMojo.java
@@ -63,6 +63,12 @@ public final class InstallNodeAndNpmMojo extends AbstractFrontendMojo {
     @Parameter(property = "skip.installnodenpm", defaultValue = "${skip.installnodenpm}")
     private boolean skip;
 
+    @Parameter(property = "numberOfRetries", defaultValue = "1")
+    private int numberOfRetries;
+
+    @Parameter(property = "intervalBetweenRetriesMs", defaultValue = "10000")
+    private int intervalBetweenRetriesMs;
+
     @Component(role = SettingsDecrypter.class)
     private SettingsDecrypter decrypter;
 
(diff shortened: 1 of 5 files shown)


melloware (approved):

melloware:
@eirslett can we get this bumped or merged.  GitHub Actions is constantly failing with this error?

eirslett:
I see the problem! It's just that I'm a bit unsure what the right solution is. There are trade-offs.

There are already so many configuration options on this plugin - and if you look at the PRs that have been opened, so many proposed configuration options - in the end it will just be even more unmaintainable than it already is. It's hard to write automated tests for everything without a major refactor of the codebase, and it's not uncommon that unrelated things break because of that.

Maybe adding dynamic retries is an option, I'm not sure it has to be configurable though. Could just go with a simple default, for example 3 retries, 10 seconds between each. But retrying against a CDN (nodejs.org in this case) that is overloaded usually won't fix the problem anyways, it could make the problem even worse.

There was a suggestion that we could stop streaming the downloaded file - instead just keep it in memory until it's fully downloaded. All or nothing. That will probably solve the issue with corrupt files.

As for Flink - Maybe Flink could host a mirror of the node.js CDN, for its use? This plugin has specific configuration for using other download roots. It's made specifically to be able to use a more reliable CDN than nodejs.org.

melloware:
I am going to keep my eye on 1.14.2 to see if it alleviates our GitHub Actions failures thanks for looking into this @eirslett

melloware:
OK getting this error now...

```
Failed to execute goal com.github.eirslett:frontend-maven-plugin:1.14.2:install-node-and-npm (install node and npm) on project primefaces: 
Could not download Node.js: Could not download https://nodejs.org/dist/v18.8.0/node-v18.8.0-linux-x64.tar.gz: Premature end of Content-Length delimited message body (expected: 43,238,752; received: 11,943,636) ->
```

eirslett:
> OK getting this error now...

How often is this occurring - 100 % of the time, or is it only when the Node.js CDN is unstable?

melloware:
Its seems really random.   So it must just be when Node.js CDN is busy?

dennisoelkers:
@eirslett: are there any objections against merging this? We are also facing the premature EOF issue on github actions every day and think this could help us improving it.

PR #1116, opened 2023-10-23
Use a script to proxy PNPM executable on Windows

### LINKED ISSUE #1115: PNPM executable not exposed on Windows
This issue is equivalent to #967, but exclusively for for Windows. The decision was made to fix this issue for all other operating systems, [but not for Windows](https://github.com/eirslett/frontend-maven-plugin/pull/1045#issuecomment-1272384510), due to problems with symlinks on that platform.

jonkoops:
Since Windows does not support symbolic links (or at least not consistently), the choice was made [not to support](https://github.com/eirslett/frontend-maven-plugin/pull/1045#issuecomment-1272384510) the PNPM executable under Windows. This PR adds support for running the PNPM executable by adding a script under the same name in the installation directory, and proxy any commands sent to it to the PNPM executable.

This also allow the removal of some of the testing profiles that were previously added as the tests would not pass under Windows. And also removes the legacy linking code that was copied from the NPM version that never worked.

Closes #1115

### DIFF
--- a/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/PnpmInstaller.java
+++ b/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/PnpmInstaller.java
@@ -1,15 +1,14 @@
 package com.github.eirslett.maven.plugins.frontend.lib;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
-import java.io.EOFException;
-import java.io.File;
-import java.io.IOException;
-import java.nio.file.Files;
-import java.util.Arrays;
-import java.util.HashMap;
 import org.apache.commons.io.FileUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.io.*;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.HashMap;
 
 public class PnpmInstaller {
 
@@ -69,7 +68,12 @@ public class PnpmInstaller {
             if (!pnpmIsAlreadyInstalled()) {
                 installPnpm();
             }
-            copyPnpmScripts();
+
+            if (this.config.getPlatform().isWindows()) {
+                linkExecutableWindows();
+            } else {
+                linkExecutable();
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/PnpmInstaller.java
```diff
@@ -169,42 +173,72 @@ private void installPnpm() throws InstallationException {
...
-                    {
-                        FileUtils.copyFile(scriptFile, copy);
-                    }
-                    catch (IOException e)
-                    {
-                        throw new InstallationException("Could not copy pnpm", e);
-                    }
-                    copy.setExecutable(true);
-                }
-            }
```
jonkoops, line 193: I have removed this old code as it seems to have been copied from the `NpmInstaller`, but never actually functioned. I've manually checked every release of PNPM, and none of them include a `pnpm` or `pnpm.cmd` file in their `bin` directory.

frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/PnpmInstaller.java
```diff
@@ -169,42 +173,72 @@ private void installPnpm() throws InstallationException {
...
+                .append(":: Created by frontend-maven-plugin, please don't edit manually.\r\n")
+                .append("@ECHO OFF\r\n")
+                .append("\r\n")
+                .append("SETLOCAL\r\n")
+                .append("\r\n")
+                .append(String.format("SET \"NODE_EXE=%%~dp0\\%s\"\r\n", relativeNodePath))
+                .append(String.format("SET \"PNPM_CLI_JS=%%~dp0\\%s\"\r\n", relativePnpmPath))
+                .append("\r\n")
+                .append("\"%NODE_EXE%\" \"%PNPM_CLI_JS%\" %*")
+                .toString();
```
jonkoops, line 234: This script is partially taken from the [NPM equivalent](https://github.com/npm/cli/blob/c422a01e1564d25148c821ee257196ebe60e8e6c/bin/npm.cmd), but pruned down to the bare essentials needed for this use-case.

jonkoops:
Looks like the tests on Windows are flaking out on something unrelated to the changes in this PR? Perhaps a re-run will pass them.

eirslett:
Looks like it works. Thanks!

jonkoops:
Thanks for getting to this so quick @eirslett, much appreciated.

eirslett:
Could you please help update the changelog as well? I think this will be released as a patch version.

jonkoops:
Sure, opened a PR under #1117

PR #1119, opened 2023-10-26
Verify checksum for downloaded Node.js archive (fixes #645)

### LINKED ISSUE #645: Verify the content of downloaded node dist
Multiple times I have seen a partial version of node installed in {local-repo}/com/github/eirslett/node/... both on Linux and Windows. Removing this file usually fixes the problem, but it is happening more and more frequently. 

What I am experiencing is that a build just hangs forever (I just had one hung for 2 days). The last logs I see are this:
```
[INFO] Downloading https://nodejs.org/dist/v6.11.2/node-v6.11.2-linux-x64.tar.gz to /lcl/dev/data/builds/appsTools/jobs/gfs-boot Sonar/workspace/gfs-boot-starters/gfs-boot-starter-web/target/local-repo/com/github/eirslett/node/6.11.2/node-6.11.2-linux-x64.tar.gz
[INFO] No proxies configured
[INFO] No proxy was configured, downloading directly
```
This is immediately after a successful build and then cleaning out the local Maven artifact repo. 

On linux, what I saw this last time in the local repo is a node-v6.11.2-linux-x64.tar.gz file that is about 2.9 MB, not the 13.1 MB that should be there. Looking at the contents of the file, it all looks correct other than it is truncated at 2.9 MB. I'm looking at the code a bit, but didn't dig too far, but I am not sure what would cause a partial file to be downloaded. Does the CloseableHttpResponse start writing to the .tar.gz file before it has finished downloading the content? 

I am asking for:

1. Provide a timeout so that the download of node is aborted if the network hangs
2. Download the file to a temp file in the target directory
3. Verify the contents of the file
4. Rename the temp file once complete

I see node provides a SHASUM256.txt.asc files that could be used to verify the contents. 

There is no proxy in the way. 

Bug

Build hangs

Probably the easiest way to reproduce would be to point to an alternate URL of an HTTP server that is configured to simply hang forever.

Abort the build after a timeout and not leave behind unverified node binaries that cause future builds to fail. 

frontend-maven-plugin version 1.5
Linux 3.10.0-514.21.2.el7.x86_64 #1 SMP Sun May 28 17:08:21 EDT 2017 x86_64 x86_64 x86_64 GNU/Linux

AndreasHae:
**Summary**

As explained in #645, we should verify downloaded archives to ensure its integrity to 1. mitigate issues when the file has been corrupted and 2. to prevent security issues when the file has been tampered with.

**Tests and Documentation**

I wrote two unit tests for verifying checksums. I ran the integration tests to make sure not to introduce a breaking change.

By the way, not related to this PR, but I noticed the file structure could use some cleanup as there are quite a lot of classes accumulating in one package. Is this something you would be open to @eirslett?


eirslett-visma:
I considered this approach before (in my head) but never wrote the code for it.

I think nr. 1 (corrupted downloads) are finally solved with https://github.com/eirslett/frontend-maven-plugin/pull/1118 which is in the latest version of the plugin (1.4.2). I think the chance of the Node.js CDN serving a full binary response with 200 OK header - but with a corrupt file - is very low. The main issue before was that files were only half-downloaded, because of bottlenecks on the CDN and timeouts.

Regarding issue nr. 2, the main issue with this PR is that it downloads the shasums from the same CDN that it downloads Node from. So anybody who tampered with the Node binaries could easily tamper with the SHASUM file as well, to fool the verification check. The SHASUMS from nodejs.org are signed with GPG keys, but the GPG signature is not verified. For that, you would need an up-to-date way of verifying the "official" Node.js GPG key, which could be updated/replaced in case of security breaches, or just replaced regularly as part of a security regime. You see how this is becoming a real rabbit hole?

At that point, I think a better security measure would be to host your own private asset repository (Like a Nexus instance for example) and then fetch the binaries from there instead of from nodejs.org.

AndreasHae:
I understand. Would you be open to me exploring the option to verify GPG signatures as well, or do you think the maintenance burden would be too high?

eirslett:
I think it's out of scope for this plugin. Maybe a separate Maven plugin for verifying shasums?
Once you dive into this - remember that your whole .m2 folder also consists of stuff that is downloaded from (probably) Maven Central. It would make sense to check shasums for those to as well, right?
For example for spring-core: https://repo1.maven.org/maven2/org/springframework/spring-core/6.0.13/spring-core-6.0.13.jar.sha1
But to make sure that the shasum can be trusted, you also have to check the GPG key: https://repo1.maven.org/maven2/org/springframework/spring-core/6.0.13/spring-core-6.0.13.jar.asc

I'm sure 99 % of organizations don't actually verify trust in their dependencies. That's what makes the software world so vulnerable to supply chain attacks. Maven itself (as far as I know) doesn't come with any built-in signature checks at all. But there's a plugin for it: https://www.simplify4u.org/pgpverify-maven-plugin/ maybe you could fork it, and make it work with node/npm downloads as well.

Issue #1124, opened 2023-11-23
Node cannot be installed to relative path


mvilliger:
**Do you want to request a _feature_ or report a _bug_?**
Report a Bug

**What is the current behavior?**
Using a relative path in `installDirectory` for goal `install-node-and-npm` always leads to "Bad zip entry" exception:
```
<plugin>
  <groupId>com.github.eirslett</groupId>
  <artifactId>frontend-maven-plugin</artifactId>
  <configuration>
    <installDirectory>${project.build.outputDirectory}/subfolder/../node</installDirectory>
    <nodeVersion>whatever</nodeVersion>
  </configuration>
</plugin>
```

The reason seems to be [here](https://github.com/eirslett/frontend-maven-plugin/blob/master/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/ArchiveExtractor.java#L82) where a normalized `destPath` is compared against a non-normalized `destinationDirectory`.
I assume the `destinationDirectory` should be normalized as well before comparing.

**If the current behavior is a bug, please provide the steps to reproduce.**
Execute the `install-node-and-npm` goal with a relative `installDirectory` for node as shown in the sample above.

**What is the expected behavior?**
Node is correctly extracted to the normalized path of the given relative `installDirectory`.

**Please mention your frontend-maven-plugin and operating system version.**
frontend-maven-plugin version: 1.14.2
operating system: any

eirslett:
What is the use case for this? Is it not possible to achieve using `${maven.multiModuleProjectDirectory}` or `${project.parent.basedir}` or something like that? Why is node being installed inside `outputDirectory` (I assume it's one of the `target` folders)?

mvilliger:
We use a general parent pom which installs Node into the parent module's target dir as soon as the first sub-module is built which requires Node. For this the destination directory is configured using various properties which some might contain `/../`.

Anyway it should not have any downside to normalize both directories as comparing them only makes sense if both are compared normalized. Don't you think?

mvilliger:
Another point: These kind of installDirectories used to work finde in version < 1.13. So it seems like a regression concerning this feature. Currently, this blocks us from updating the plugin.

PR #1108, opened 2023-09-15
Basic bun integration

deemkeen:
**Summary**

Basic bun integration as requested in #1106
Bun still doesn't work with windows, so i've tried to skip the windows integration tests, maybe there is a better way to do it

PR #1125, opened 2023-11-23
Normalize the destination directory when extracting zip files

mvilliger:
Otherwise, relative installDirectory paths cannot be used because the zip-slip check fails with "Bad zip entry" exception even the zip is fine.

Closes https://github.com/eirslett/frontend-maven-plugin/issues/1124


eirslett:
ok let's merge it!

mvilliger:
Thanks :)
Any plans when this will be available on Maven central?

eirslett:
I think I will try to get https://github.com/eirslett/frontend-maven-plugin/pull/1108 merged first.

PR #1130, opened 2024-01-04
Fix `<configuration>` nesting in maven examples

viliam-durina:
**Summary**

**Tests and Documentation**


liefke:
@viliam-durina: That is not true - you can add a `<configuration>` element to the `<plugin>` section, which applies to all `<execution>` sections then.

PR #1143, opened 2024-04-29
fix: yarn berry detection in case working directory is changed

holomekc:
**Summary**
In case the workingDirectory option is used the detection of yarn berry does not work properly:
image

To fix that the workingDirectory needs to be included for the .yarnrc.yml file detection

As a workaround it is possible to add an empty .yarnrc.yml file into the module root directory.
image

Furthermore, I forwarded the is berry check to the runner, so that additional arguments are not set, which are not supported by yarn berry.

**Tests and Documentation**
I updated the workflows, because they complained because of outdated and not supported node versions and updated the java version, because the mac os run complained that Java 8 is not supported anymore.

Furthermore, I added a test for yarn berry.


.github/workflows/main.yml
```diff
@@ -13,18 +13,18 @@ jobs:
...
           path: ~/.m2
           key: "${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}"
           restore-keys: |
             ${{ runner.os }}-maven3-
       - name: Configure Java
-        uses: actions/setup-java@v3
+        uses: actions/setup-java@v4
         with:
-          java-version: "8"
+          java-version: "21"
```
eirslett, line 27:
```suggestion
          java-version: "8"
```

We try to support Java 8 still (unless there is a good reason not to)

.github/workflows/main.yml
```diff
@@ -13,16 +13,16 @@ jobs:
...
         with:
           path: ~/.m2
           key: "${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}"
           restore-keys: |
             ${{ runner.os }}-maven3-
       - name: Configure Java
-        uses: actions/setup-java@v3
+        uses: actions/setup-java@v4
         with:
           java-version: "8"
```
eirslett, line 27:
```suggestion
          java-version: "11"
```

Seems like due to a newer MacOS version on GitHub Actions, there's no Java 8, so probably time to bump to Java 11 then.

eirslett:
Thanks a lot!

PR #1153, opened 2024-06-07
Fix #1152: Copy npm script file

### LINKED ISSUE #1152: NodeInstaller is not copying the npm binaries next to the node exec
This causes `package.json` script using `npm` to yield: `npm command not found`, because it is not in the PATH.

There is a comment saying it does, but it's is not actually done:
https://github.com/eirslett/frontend-maven-plugin/blob/master/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NodeInstaller.java#L192

It works when using the NPMInstaller because it actually does it:
https://github.com/eirslett/frontend-maven-plugin/blob/master/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NPMInstaller.java#L182-L205

melloware:
**Summary**

Fix #1152: Copy npm script file

As mentioned in the ticket the comment said it was copying the script and it was not.  

Originally reported on Quarkus Quinoa: https://github.com/quarkiverse/quarkus-quinoa/issues/683

### DIFF
--- a/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NodeInstaller.java
+++ b/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/NodeInstaller.java
@@ -193,7 +193,18 @@ public class NodeInstaller {
                     for (String script : Arrays.asList("npm", "npm.cmd")) {
                         File scriptFile = new File(npmDirectory, "bin" + File.separator + script);
                         if (scriptFile.exists()) {
-                            scriptFile.setExecutable(true);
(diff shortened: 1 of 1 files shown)


eirslett:
Should it be copying "npx" and "npx.cmd" as well?

melloware:
Good question i only enhanced what was there i think this code is in two places in your code base.

melloware:
Updated to copy "npx" and "npx.cmd" as well

melloware:
Bump

eirslett:
Ah yeah, sorry, thanks!

melloware:
Thank you!

Issue #1032, opened 2022-04-16
Support for corepack


timja:
**Do you want to request a _feature_ or report a _bug_?**
Feature

The preferred way to manage the version of yarn is now corepack not checking the yarn file into the git repo: https://yarnpkg.com/getting-started/install

**What is the current behavior?**
No support

**What is the expected behavior?**

- A flag to enable [corepack](https://nodejs.org/dist/latest/docs/api/corepack.html) support
- The plugin should run `corepack enable`
- Don't require setting a `yarn` version as it will be set in the package.json and corepack will handle fetching it

**Please mention your frontend-maven-plugin and operating system version.**
-

long76:
up vote here

bencope:
@eirslett any chance of having this one merged?

PR #1157, opened 2024-06-12
Support corepack

stevestorey:
**Summary**

Adds support for using `corepack` to manage the package manager in use on a given build project.

**Tests and Documentation**

README is updated, an integration example has been provided

Fixes: https://github.com/eirslett/frontend-maven-plugin/issues/1032

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,9 +56,11 @@ to see how it should be set up: https://github.com/eirslett/frontend-maven-plugi
 
  - [Installing node and npm](#installing-node-and-npm)
  - [Installing node and yarn](#installing-node-and-yarn)
+ - [Installing node and corepack](#installing-node-and-corepack)
  - Running 
     - [npm](#running-npm)
     - [yarn](#running-yarn)
+    - [corepack](#running-corepack)
     - [bower](#running-bower)
     - [grunt](#running-grunt)
     - [gulp](#running-gulp)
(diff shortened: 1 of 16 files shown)


ansgarkroger:
@eirslett, do you think there's a chance to get this PR merged and released? Corepack seems to be the best way to support modern Yarn (only way I got it to work in my project)

davidmedinasigasi:
@eirslett, any change this could get merged? my team would really benefit from this

stevestorey:
I've added one more commit to use the corepack version provided with the Node install if no override version is provided (which is probably what most people want to do)

eirslett:
Thanks! Looks like this is a much requested feature, so I'll merge it in! (given that the test suite works)

benkelaar:
@eirslett Anything I can do to help get this released?

I was slightly thrown off by the corepack targets in the readme that didn't seem to work :-)

eirslett:
Hey! Sorry, I tried to release it but I got some auth issues from Sonatype. It will probably take a few hours of digging :-/
And I was a bit surprised by the news that Node.js will (probably?) stop bundling Coreback. So maybe this plugin shouldn't support using the bundled one.

stevestorey:
> And I was a bit surprised by the news that Node.js will (probably?) stop bundling Coreback

Thanks for highlighting. That's very disappointing :( - tho as yet they've not put a timeframe on dropping it from the bundle.

> maybe this plugin shouldn't support using the bundled one.

I would suggest that until they do set a timeframe, we can safely continue to use the bundled version.

Since bumping the node version on the plugin is an explicit step, even if we changed nothing and a new version of Node is published without corepack - nothing should break for existing users' builds.

When they bump their node version on the plugin, they'll pretty quickly realise they need to set a corepack version as well (which, from what I read, _should_ still work and be a supported config).

PR #1165, opened 2024-09-15
#1160 Property to select node path.

### LINKED ISSUE #1160: How about accept a property about to the node installiation in system for supporting some arch or custom patch?
It's a feature.  

I am building packages of some software for Arch Linux RISC-V Port.  
Node.js did not provide the prebuilt binary for RISC-V, but we have a patched one provided by the system's package manager.  
Jenkins used frontend-maven-plugin to build their frontend page, we can't do many patches to the source in order to build it, so I want to make it use the node in the system.  
Would you be willing to accept a property point to a node executable file to replace the default downloaded one? Also, users no longer need to use the exec-maven-plugin.  
If that sounds good, I'd love to create a PR for it.

qyl27:
**Summary**

It was added a way to use the existing nodejs installation which was in the system.  
(Closes #1160 .)

**Tests and Documentation**

I have added a document line for it.  
And I have tested it by using a minimum project example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ to see how it should be set up: https://github.com/eirslett/frontend-maven-plugi
     - [Ignoring Failure](#ignoring-failure)
     - [Skipping Execution](#skipping-execution)
     
+
 **Recommendation:** _Try to run all your tasks via npm scripts instead of running bower, grunt, gulp etc. directly._
 
 ### Installing node and npm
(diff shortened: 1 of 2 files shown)


qyl27:
@eirslett

Aleod-m:
+1 would be very usefull on nixos as on nixos we cannot run dynamicly linked executables so i would provide the node from the nixpkgs

Stepiiis:
Hi, are there any updates on this? Currently trying to build on NixOS and failing because I was not able find a way to use local node.

qyl27:
> Hi, are there any updates on this? Currently trying to build on NixOS and failing because I was not able find a way to use local node.

Hello, this PR is ready and works well in ArchLinux RISC-V port packaging.

PR #1171, opened 2025-01-13
Avoid modification of https.protocols

### LINKED ISSUE #1170: FileDownloader disables TLSv1.3

Bug

[FileDownloader restricts TLS to TLSv1.2](https://github.com/eirslett/frontend-maven-plugin/blob/640d6f065435102ea0c3580416d1653c6d75c767/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/FileDownloader.java#L59).

This is a problem for setups that only allow TLSv1.3. Even worse, it doesn't just restrict the TLS version this plugin uses, but also all plugins that run after it in the same JVM instance, because it modifies global properties.

Connections to TLSv1.3 only hosts will fail with `Received fatal alert: protocol_version`.

Use the plugin with a host that only supports TLSv1.3 or use a plugin that makes HTTPS requests to a TLSv1.3 only host after running this plugin and downloading a file with `FileDownloader`.

TLSv1.3 should be used.

1.15.1 / Debian 11

kelunik:
This disables TLSv1.3 and changes global state that shouldn't be changed.

Fixes #1170.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,13 +18,12 @@ install:
   - cmd: SET M2_HOME=C:\maven\apache-maven-%MAVEN_VERSION%
   # Prepend Java entry, remove Ruby entry (C:\Ruby193\bin;) from PATH
   - cmd: SET PATH=%M2_HOME%\bin;%JAVA_HOME%\bin;%PATH:C:\Ruby193\bin;=%;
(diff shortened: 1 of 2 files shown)


kelunik:
The AppVeyor build doesn't seem active in this PR, so not sure whether it still works without the config option or not.

pitgrap:
LGTM. Did you test it locally with your TLS1.3 setup?

PR #1197, opened 2025-12-16
Update Dependencies

michaelmejaeger:
This PR is for updating Maven dependencies and setting the Java-Version from 1.8 to 17 as discussed

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,8 +9,6 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <java.version>17</java.version>
-        <maven.compiler.source>17</maven.compiler.source>
-        <maven.compiler.target>17</maven.compiler.target>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -8,9 +8,9 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <java.version>1.8</java.version>
-        <maven.compiler.source>1.8</maven.compiler.source>
-        <maven.compiler.target>1.8</maven.compiler.target>
+        <java.version>17</java.version>
+        <maven.compiler.source>17</maven.compiler.source>
```
sepe81: Just a small idea: maybe integrating maven.compiler.release would be an even better fit.

See https://docs.openrewrite.org/recipes/java/migrate/maven/usemavencompilerpluginreleaseconfiguration

Issue #768, opened 2018-08-30
feature request: add flag to skip install-node-and-npm and install-node-and-yarn


philipreimer:
**What is the current behavior?**
Currently, there are flags like `-Dskip.npm` to skip a tool's execution. But there does not seem to be a flag to also skip the installation of the respective tool, e.g. to skip the goal `install-node-and-npm`.

Hence, even if npm execution is skipped, node and npm are still installed, although they are not needed.

**What is the expected behavior?**
There should be a flag like `-Dskip.node-and-npm.install` to skip the actual installation goal.

It should be possible to specify this flag both on the command line via `-Dskip.node-and-npm.install` and in the plugin's execution configuration like:
```
<execution>
    <id>install node and npm</id>
    <goals>
        <goal>install-node-and-npm</goal>
    </goals>
    <phase>...</phase>
    <configuration>
        <skip.node-and-npm.install>true</skip.node-and-npm.install>
        <nodeVersion>...</nodeVersion>
        <npmVersion>...</npmVersion>
    </configuration>
</execution>
```

**Please mention your frontend-maven-plugin and operating system version.**
frontend-maven-plugin version 1.6
Maven 3.5.2
OpenJDK 1.8.0_181
Antergos Linux

philipreimer:
D'oh! I just noticed that the configuration option `<skip>` already exists.
```
<execution>
    <id>install node and npm</id>
    <goals>
        <goal>install-node-and-npm</goal>
    </goals>
    <phase>...</phase>
    <configuration>
        <skip>true</skip>
        <nodeVersion>...</nodeVersion>
        <npmVersion>...</npmVersion>
    </configuration>
</execution>
```

jamesrwoodruff:
Good to know the skip configuration option is available. But per your original question, can the skip be invoked from the command line (e.g., -D.skip.install-node-and-npm)?

jamesrwoodruff:
Actually, I found it in InstallNodeAndNpmMojo.java: -Dskip.installnodenpm

Issue #758, opened 2018-08-10
Question: how to skip yarn and npm download and install


huksley:
When i run mvn -Dskip.yarn=true -Dskip.npm=true npm and yarn downloaded and installed.
How to disable this?

My config is the following:
```xml
<plugin>
    <groupId>com.github.eirslett</groupId>
    <artifactId>frontend-maven-plugin</artifactId>
    <version>1.6</version>
    <configuration>
        <installDirectory>target</installDirectory>
        <workingDirectory>src/main/resources/frontend</workingDirectory>
        <nodeVersion>v8.11.1</nodeVersion>
        <yarnVersion>v1.7.0</yarnVersion>
    </configuration>
    <executions>
        <execution>
            <id>install-node-and-yarn</id>
            <goals>
                <goal>install-node-and-yarn</goal>
            </goals>
            <phase>generate-resources</phase>
        </execution>
        <execution>
            <id>yarn-install</id>
            <goals>
                <goal>yarn</goal>
            </goals>
            <phase>generate-resources</phase>
            <configuration>
                <arguments>install</arguments>
            </configuration>
        </execution>
        <execution>
            <id>yarn-build</id>
            <goals>
                <goal>yarn</goal>
            </goals>
            <phase>generate-resources</phase>
            <configuration>
                <arguments>run build</arguments>
                <environmentVariables>
                    <!-- FIXME: disable service workers during build -->
                    <REACT_APP_OFFLINE_FIRST>false</REACT_APP_OFFLINE_FIRST>
                </environmentVariables>
            </configuration>
        </execution>
    </executions>
</plugin>
```

thu:
I think you can not install it on target folder.

php-coder:
I've just found in the sources that it's possible by specifying `-Dskip.installyarn` and `-Dskip.installnodenpm` arguments.

But:
1) noth option aren't documented in `README.md`
2) https://github.com/eirslett/frontend-maven-plugin/wiki mentions only one of the options

So, now this is rather a documentation issue.

PR #1203, opened 2026-02-05
[doc] Add missing skip.XXX documentation to the README.md

yonatang:
**Summary**
Adding undocumented `skip.XXX` options, such as `skip.installnodenpm`.

It seems like a few (me included) were looking for these options: 
* https://github.com/eirslett/frontend-maven-plugin/issues/758
* https://github.com/eirslett/frontend-maven-plugin/issues/768
* probably more

If that feature exists, there is no reason to keep it in the dark, and furthermore, making LLMs take a wild guesses how to do that.

**Tests and Documentation**

Just documentation - README.md was updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -716,10 +716,14 @@ Build tools and property to enable skipping:
 - yarn `-Dskip.yarn`
 - bower `-Dskip.bower`
 - bun `-Dskip.bun`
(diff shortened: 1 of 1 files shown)


eirslett:
Thanks!

PR #1225, opened 2026-05-21
use .mjs file for pnpm if it exists

### LINKED ISSUE #1224: pnpm install no longer works with 11+
The .cjs file in the distribution is no longer marked as executable. It's expected that the .mjs file will be used instead. 

I was able to manually workaround this in https://github.com/keycloak/keycloak/pull/48975 by changing the file permissions.

shawkins:
closes: #1224 - the pnpm 11+ kit does not include an executable .cjs file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Last public release: [![Maven Central](https://maven-badges.herokuapp.com/maven-
 
 ## Changelog
 
+### 2.0.1
+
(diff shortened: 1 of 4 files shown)


eirslett:
Thanks a lot!

alebx:
@eirslett when will this fix/feature be released?

eirslett:
Should be ready now! @alebx

Issue #1232, opened 2026-07-02
Add support for LoongArch64 (loong64) architecture


xuxiaowei-com-cn:
### Background

LoongArch64 is a CPU architecture independently developed in China (Loongson architecture). Node.js `linux-loong64` binary builds are maintained by the [nodejs/unofficial-builds](https://github.com/nodejs/unofficial-builds) project and distributed at https://unofficial-builds.nodejs.org/download/release/ .

According to an actual scan (`find_loong64_versions.py`), linux-loong64 builds cover the following versions:
- **Earliest**: v18.18.0
- **Latest**: v26.4.0
- 87 versions in total, spanning v18, v20, v21, v22, v23, v24, v25, v26 major releases

> ⚠️ **Note**: `linux-loong64` builds are **NOT** available on the official Node.js download site `https://nodejs.org/dist/`; they are provided by `unofficial-builds.nodejs.org`.

`frontend-maven-plugin` currently does not recognize this architecture, making it impossible to automatically download the correct Node.js binary on LoongArch platforms.

### Current Behavior

1. When running on a LoongArch platform, `os.arch` returns `"loongarch64"`, but the `Architecture.guess()` method does not recognize this value and falls back to the `arch.contains("64") ? x64 : x86` logic, incorrectly matching it to `x64`.
2. Even if the architecture is recognized correctly, the default download root `https://nodejs.org/dist/` does not have `linux-loong64` builds, resulting in a 404 download failure.

### Expected Behavior

1. The plugin should correctly recognize the `loongarch64` architecture and map it to the `linux-loong64` classifier.
2. For the loong64 architecture, the plugin's default download root should point to `https://unofficial-builds.nodejs.org/download/release/` (similar to the existing musl/Alpine handling), or allow users to specify it via the `nodeDownloadRoot` configuration option.

### Scope of Changes

- `Platform.java`: Add `loong64` enum value and recognition logic for `"loongarch64"`; switch to the unofficial-builds download root for loong64 in `guess()`.
- `PlatformTest.java`: Add corresponding unit tests.

### References

- [nodejs/unofficial-builds](https://github.com/nodejs/unofficial-builds) — Repository for loong64 builds
- [Issue #97: Loong64 missing from index](https://github.com/nodejs/unofficial-builds/issues/97) — loong64 index fix (September 2023)
[...]

PR #1233, opened 2026-07-02
Add support for LoongArch64 (loong64) architecture

xuxiaowei-com-cn:
https://github.com/eirslett/frontend-maven-plugin/issues/1232

**Summary**

Add LoongArch64 (loong64) architecture support to the plugin.

**Motivation**: LoongArch64 is a CPU architecture independently developed in China (Loongson). On LoongArch platforms, Java's `os.arch` system property returns `"loongarch64"`. Currently, `Architecture.guess()` does not recognize this value and falls back to `arch.contains("64") ? x64 : x86`, incorrectly matching it to `x64` and causing the plugin to download the wrong Node.js binary.

This PR adds proper recognition so that the plugin produces the correct `linux-loong64` classifier. Node.js `linux-loong64` binaries are available from [unofficial-builds.nodejs.org](https://unofficial-builds.nodejs.org/download/release/) (earliest: v18.18.0, latest: v26.4.0, 87 versions across v18–v26).

> **Note**: `linux-loong64` builds are on `unofficial-builds.nodejs.org`, not the default `nodejs.org/dist/`. Users on LoongArch must override `nodeDownloadRoot` to `https://unofficial-builds.nodejs.org/download/release/` in plugin configuration for the download to succeed. A follow-up change could switch the download root automatically (similar to the musl/Alpine handling).

**Changes**:

| File | Change |
|------|--------|
| `Platform.java` | Add `loong64` to the `Architecture` enum; add `"loongarch64"` → `loong64` mapping in `guess()` |
| `PlatformTest.java` | Add `detect_linux_loong64` test verifying `"linux-loong64"` classifier |

**Tests and Documentation**

- **Unit test**: `detect_linux_loong64` verifies that `Platform.guess(OS.Linux, Architecture.loong64, () -> false)` resolves to classifier `"linux-loong64"`.
- The change follows the existing pattern for other architectures (ppc64le, s390x, arm64, etc.) — no new abstractions introduced.
- A one-liner add to CHANGELOG.md: "Add LoongArch64 (loong64) architecture support."

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Last public release: [![Maven Central](https://maven-badges.herokuapp.com/maven-
 
 ## Changelog
 
+### 2.0.2 (unreleased)
+
+* Add LoongArch64 (loong64) architecture support ([#1232](https://github.com/eirslett/frontend-maven-plugin/issues/1232))
+
 ### 2.0.1
 
 * Use pnpm .mjs file if it exists to support 11+ ([#1224](https://github.com/eirslett/frontend-maven-plugin/issues/1224))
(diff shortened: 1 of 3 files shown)


xuxiaowei-com-cn:
@eirslett Would you mind reviewing PR #1233 for LoongArch64 architecture support? Your feedback would be very helpful.

